PR #20, opened 2017-07-06
Parsing and display of program service description

KeyserSoze1:
Added parsing of ID3 from PSD.  Uses external id3v2lib to parse once a complete PSD has been grabbed.  ID3 tags are sent to the console output for display.


awesie:
Thanks for putting this together. I'm going to merge the pull request then make some formatting/stylistic changes in a separate commit.

PR #21, opened 2017-07-07
Fix MER calculation.

argilo:
I think the MER dB calculation is off by a factor of 2. That can be corrected by removing the square roots.


awesie:
After consulting Wikipedia (and specifically the Talk page for MER) and Stack Exchange, I am now convinced of my error. The correct definition of MER should be the ratio of average signal power / average error power, where the power of a IQ sample is I^2 + Q^2. So, no square root is needed.

PR #22, opened 2017-07-08
Fix coarse carrier frequency offset correction.

argilo:
The `find_ref` function does not take phase ambiguity into account, so it only detects reference subcarriers half the time. After I added a check for the other phase, the coarse carrier frequency offset correction went from working half the time to all the time.

I also reduced the search window to 76 subcarriers. Since there are only four possible reference subcarrier id's, it seems that the standard was only designed to account for that much frequency offset (approximately +/- 128 ppm at the upper end of the FM band).  A larger search window could result in the wrong reference subcarrier being detected (at least if the station is transmitting in the extended hybrid mode). I imagine +/- 128 ppm is sufficient for RTL-SDR dongles. The worst one I had was off by 100 ppm.

### DIFF
--- a/src/sync.c
+++ b/src/sync.c
@@ -148,6 +148,18 @@ static int find_ref (float complex *buf, unsigned int ref, unsigned int rsid)
         }
         if (i == sizeof(needle))
             return n;
+        for (i = 0; i < sizeof(needle); i++)
+        {
+            // first bit of data may be wrong, so ignore
+            if ((n + i) % BLKSZ == 0) continue;
+            // ignore don't care bits
+            if (needle[i] < 0) continue;
+            // test if bit is correct
+            if (needle[i] == data[(n + i) % BLKSZ])
+                break;
+        }
+        if (i == sizeof(needle))
+            return n;
     }
     return -1;
 }
@@ -225,7 +237,7 @@ void sync_process(sync_t *st, float complex *buffer)
         }
         else if (st->cfo_wait == 0)
         {
-            for (i = -300; i < 300; ++i)
+            for (i = -38; i < 38; ++i)
             {
                 int offset2;
                 adjust_ref(buffer, st->phases, LB_START + i + BAND_LENGTH - 1);


awesie:
Thanks for the fixes. I'll review the open pull requests this weekend.

awesie:
I'm trying to convince myself that this change is correct. I'm okay with reducing the CFO search range, I chose something large to be _safe_ and I wasn't sure what of the general range for a rtl-sdr.

The phase ambiguity concern strikes me as wrong though, because the reference subcarrier uses DBPSK. The differential encoding _should_ eliminate the phase ambiguity, except for the first bit. Thoughts?

argilo:
The same thought occurred to me this morning, and I'm now confused why this code change appears to fix the problem. Let's put this on hold until we sort out what's going on.

argilo:
I think the phase ambiguity arises from the `slope` (calculated by `calc_phase`). It seems to increase with the CFO but I'm not sure yet why that is.

argilo:
I think I understand what's going on here. If you're off in frequency by one subcarrier then each subcarrier is shifted in frequency by Δf. Since the symbol duration is (1+α)/Δf, each symbol is rotated by 1+α turns compared to the previous symbol, giving a measured slope of α turns per symbol. By the time you're off in frequency by 5 subcarriers, the slope is greater than a quarter turn per symbol and phase ambiguity of the slope manifests.

Thus it seems necessary to account for the phase ambiguity of the slope measurement. When the wrong phase is measured, the reference subcarrier bits are flipped, which my proposed change accounts for.

I'm convinced that this change is correct now, so you can merge it if you like (or change it to something better if you can see another way to do it).

argilo:
I also did some measurements and verified (to within one part per thousand or so) that the slope increases by α turns per subcarrier.

awesie:
I applied your patch, but I also took this opportunity to factor out some of the shared code for find_first_block and find_ref. Let me know if it breaks your fix.

argilo:
@awesie Everything's still looking good after your refactor. :+1:

PR #23, opened 2017-07-10
Allow SPSD ports 0x5201 to 0x5207.

argilo:
Currently, program service data is only displayed for the main program. Ports 0x5201 to 0x5207 are used for the secondary programs (HD2, HD3, etc). See table 5-2 in http://www.nrscstandards.org/sg/nrsc-5-c/1019sg.pdf

/cc @KeyserSoze1

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -326,7 +326,7 @@ void output_psd_push(uint8_t *psd, unsigned int len)
 #ifdef HAVE_ID3V2LIB
     uint16_t port = *(uint16_t *)psd;
     uint16_t seq = *(uint16_t *)(psd + 2);
-    if (port != 0x5100)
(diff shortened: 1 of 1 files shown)


KeyserSoze1:
I obviously ignored the port values for my initial request.  This is good to see the improvements made.  I had one comment I placed on your change commit that I think will potentially cause a problem.  It's a minor one.  I've never seen a comment tag on any of the local stations here.

Are you aware of how the album art is transmitted?  I suspect it's in the auxiliary data section.  Have you seen anything that indicates the how the album art is transmitted.  I know there is a tag in the psd "XHDR" that indicates some information in that regard https://hdradio.com/broadcasters/engineering-support/artist-experience but am interested to try to find and decode the images if possible.

PR #27, opened 2017-07-11
Invert spectrum to simplify receiver.

argilo:
Section 14.2.2 of http://www.nrscstandards.org/SG/NRSC-5-C/1011sG.pdf specifies that the transmitter's output signal is conjugated, inverting the frequency spectrum. Doing the same in the receiver eliminates the need to read out the bits in reverse order.


awesie:
Nice catch. When I wrote my original implementation, I completely missed that and was really confused for a while. Glad to know what the actual issue was.

PR #37, opened 2017-07-29
Handle PDUs that don't have a Header Expansion Field.

argilo:
Some stations (e.g. KQRT in Las Vegas) don't include a Header Expansion Field. Currently `find_program` returns -1 in this case, because it doesn't know what to do when the Header Expansion Flag is 0.

In that case there is only one program (number 0), so we should just check whether the user asked for program number 0.

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -227,6 +227,11 @@ static int find_program(uint8_t *buf, int program, size_t length)
             if (((buf[i] >> 4) & 7) == 1 && program == ((buf[i] >> 1) & 7))
                 return start;
         }
+        else
+        {
+            if (program == 0)
+                return start;
+        }
 
         // skip remaining bytes using last location
         i = start + parse_location(buf + start, lc_bits, hdr.nop - 1) + 1;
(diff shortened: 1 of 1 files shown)


awesie:
Fixed in 8a9c8b3.

argilo:
@awesie Your change in 8a9c8b37e161b2600fa6e949ea97c800a0aaf691 doesn't work properly for stations using primary mode 3. If you ask for program 0, you get both program 0 from the P1 sidebands and the first program from the P3 sidebands (in my case, program 2). As a result of the two programs getting mixed together, you get no audio and lots of this:

`ERROR output.c:130: Decode error: Unexpected channel configuration change`

awesie:
Ah, thank you. Is it guaranteed that P3 will have a HEF?

argilo:
I imagine so, since the HEF is necessary to have more than one program.

argilo:
I added a revert of 8a9c8b37e161b2600fa6e949ea97c800a0aaf691 into this PR, so you can merge it if you're happy with my version (which I've tested on various mode 1 & mode 3 stations).

PR #39, opened 2017-08-02
Add necessary Debian packages.

argilo:
I did a minimal Debian 9 netinst to check what packages are actually necessary to build, and updated the list accordingly.


tylert:
@argilo LGTM

PR #48, opened 2017-08-17
Correct the pulse shaping window function.

argilo:
The current pulse shaping window function is raised to the power 2. The one specified in http://www.nrscstandards.org/SG/NRSC-5-C/2646s02.pdf Appendix A is not squared. I don't believe the squaring is necessary, because the transmitter also has its own pulse shaping window function (http://www.nrscstandards.org/SG/NRSC-5-C/1011sG.pdf section 13.2) and the combined effect of the two filters produces the desired squaring.

I tested this against my nrsc-5 transmitter and removing the extra squaring improved the MER by about 1 dB from ~21 to ~22 dB.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -180,15 +180,13 @@ void acquire_init(acquire_t *st, input_t *input)
     st->shape = malloc(sizeof(float) * FFTCP);
     for (i = 0; i < FFTCP; ++i)
     {
-        // The first CP samples overlap with last CP samples. Due to ISI, we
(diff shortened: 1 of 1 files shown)


awesie:
I agree, it looks like it slightly improves performance.

The section 13.2 description had confused me a couple of times, so I eventually gave up on it and just did something reasonable given the properties of the cyclic prefix. The Appendix A documentation you cite makes it very clear what we should be doing.

PR #66, opened 2017-09-22
Add support for lack of `clock_gettime` on MacOS

docprofsky:


argilo:
Thanks @docprofsky!

It looks like `clock_gettime` was only added in macOS 10.12, so a change is necessary to support older versions. Maybe it would be simpler to just use `gettimeofday`, which appears to be portable? It doesn't look like nanosecond resolution is required here.

argilo:
I'm proposing the `gettimeofday` fix in #69, so closing this PR for now.

PR #69, opened 2017-09-23
Replace clock_gettime with more portable gettimeofday.

argilo:
@docprofsky Here's a simpler fix for the OS X compile issue you ran into. Does this work for you?


docprofsky:
This works for me.

PR #70, opened 2017-09-23
Integrate with Travis CI.

argilo:
@awesie What do you think about adding Travis CI? I tried this in my fork, and it works well: https://travis-ci.org/argilo/nrsc5

I've configured it to build on Linux & OS X using both gcc and clang. The `script` also runs nrsc5 to make sure that it's able to decode your sample file. A more extensive test script could be added in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: c
+
+os:
+  - linux
+  - osx
+
+compiler:
+  - gcc
+  - clang
+
+before_install:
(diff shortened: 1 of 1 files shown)


argilo:
@awseie If you want to try this out, you'll need to sign up for Travis and give it permission to set up webhooks on the GitHub repo.

argilo:
I'm going to merge this since it makes testing easier for me. Even if you decide not to enable Travis on theori-io/nrsc5, others can use it on their forks if they like.

PR #71, opened 2017-09-23
Remove id3v2lib dependency.

argilo:
ID3v2 is not a complicated format, and HD Radio only uses a subset of it. So it seems like it's worth removing the id3v2lib dependency and implementing the parts of ID3v2 that are needed right in nrsc5. This PR implements enough to decode title, artist and album. I tested on a bunch of stations and it seems to work properly.

/cc @KeyserSoze1

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -320,15 +320,26 @@ void output_init_live(output_t *st)
 
 static unsigned int id3_length(uint8_t *buf)
 {
-    return (buf[0] << 21) | (buf[1] << 14) | (buf[2] << 7) | buf[3];
+    return ((buf[0] & 0x7f) << 21) | ((buf[1] & 0x7f) << 14) | ((buf[2] & 0x7f) << 7) | (buf[3] & 0x7f);
 }
 
 static char *id3_text(uint8_t *buf, unsigned int frame_len)
 {
-    char *text = (char *) malloc(frame_len);
-    memcpy(text, buf + 1, frame_len - 1);
-    text[frame_len - 1] = 0;
-    return text;
+    char *text;
+
+    if (frame_len == 0)
+    {
+        text = (char *) malloc(1);
+        text[0] = 0;
+        return text;
+    }
+    else
+    {
+        text = (char *) malloc(frame_len);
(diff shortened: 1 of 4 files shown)


KeyserSoze1:
No arguments from me.  As I recall there were only half a dozen fields that were allowed.  I don't even remember album being on that list.  I seem to recall a custom field being used I suspect to indicate which album art to use.  But overall I agree.  id3v2lib is probably overkill for what's needed here.

argilo:
Album (TALB) is on the list of allowed tags, and I have seen several stations using it. Album art is referenced in an XHDR tag. I found some details about that here: http://www.nrscstandards.org/SG%20ref%20docs/HD%20Radio%20Artist%20Experience%20Basics_HD%20Radio%20Alliance%20White%20Paper.pdf

src/output.c
```diff
@@ -322,65 +318,54 @@ void output_init_live(output_t *st)
...
-
-    ID3v2_frame_text_content *content = parse_text_frame_content(frame);
-
-    char temp[content->size + 1];
-    memcpy(temp, content->data, content->size);
-    temp[content->size] = '\0';
-    log_info("%s: %s", header, temp);
-
-    free(content);
+    return (buf[0] << 21) | (buf[1] << 14) | (buf[2] << 7) | buf[3];
```
awesie: Do we need to verify that these bytes are 7-bit (e.g. high bit is not set)?

src/output.c
```diff
@@ -322,65 +318,54 @@ void output_init_live(output_t *st)
...
-    ID3v2_frame_comment_content *content = parse_comment_frame_content(frame);
-
-    char temp[content->text->size + 1];
-    memcpy(temp, content->text->data, content->text->size);
-    temp[content->text->size] = '\0';
-    log_info("%s: %s", header, temp);
-
-    free(content);
+    char *text = (char *) malloc(frame_len);
+    memcpy(text, buf + 1, frame_len - 1);
```
awesie: Probably want to check if frame_len is 0

argilo:
Good idea, I'll amend this with those changes.

argilo:
Done.

PR #75, opened 2017-09-25
Wait for P3 interleaver internal matrix to fill up.

argilo:
The convolutional encoder for P3 doesn't start producing valid output until it's seen two full frames of input. To account for this, I've modified `decode_process_p3` so it only starts producing output once the interleaver's internal matrix is filled up. This should reduce the risk of crashes or other bad behaviour during startup.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -148,11 +148,17 @@ void decode_process_p3(decode_t *st)
         st->internal_p3[st->i_p3] = st->buffer_px1[i];
         (st->i_p3)++;
     }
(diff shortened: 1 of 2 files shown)


argilo:
Oops, my original commit was updating `ready_p3` too soon, and `valgrind` showed that I was still reading uninitialized memory from the internal matrix. I've changed the code to update `ready_p3` last so that `frame_push` only executes after two full frames have been processed, filling the internal matrix completely.

PR #76, opened 2017-09-25
Delay processing of AAS PDUs until first HDLC flag is seen.

argilo:
Currently, when nrsc5 starts receiving it logs several "psd crc mismatch" messages. This occurs because neither `parse_psd` nor `process_fixed_block` waits for an HDLC flag (0x7E) before starting up. As a result, they pass partial PDUs to `aas_push`.

To correct this problem, I've changed them both to use an index of -1 to indicate that the first HDLC flag has not yet been seen. Once an HDLC flag is seen, they set their index to 0 and carry on with normal processing.

It's a bit easier to see what I've changed with whitespace changes excluded: https://github.com/theori-io/nrsc5/compare/fix-aas-start?w=1

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -288,13 +288,20 @@ static void parse_psd(frame_t *st, uint8_t *data, size_t length)
     {
         index = p - data;
 
-        // complete current psd frame
-        if (st->psd_idx)
+        if (st->psd_idx >= 0)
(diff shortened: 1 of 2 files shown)


argilo:
Oops, I meant to select the experimental branch.

Issue #60, opened 2017-09-06
Better version management in brew and also version identification


classicjazz:
I have used homebrew for many years though I haven't used alternative --HEAD.

It appears that brew doesn't update existing nrsc5 installations.  It kept saying that  HEAD-320784d
was already installed.  You must uninstall nsrc5, then reinstall it. Then it rebuilds it and reflects, HEAD-1bd3f79.

Separately, could you please add a --version option to nrsc5? That would permit users to quickly confirm that they are using the latest release.

PR #78, opened 2017-09-26
Add -v option that prints the git revision.

argilo:
This partially addresses #60.

/cc @classicjazz


classicjazz:
Nice. Thank you!

Issue #81, opened 2017-09-30
Auto Gain may not work for some stations.


pclov3r:
So I noticed that the NPR station for my area the auto-gain feature does not work. It will increase it to max gain and then fail to acquire the channel.

If I set the gain manually to say 20.7 dB it works fine.

I think this may be caused by the NPR station not having very high FM deviation levels often. 

Here is a screenshot of the FFT. 
![sdr console_2017-09-30_15-04-10](https://user-images.githubusercontent.com/10610227/31049852-73b61bc6-a5f1-11e7-8d53-b059473a4fcf.png)

pclov3r:
It did use to work before with the old method so I think https://github.com/theori-io/nrsc5/pull/65 may have caused the issue.

It seems all other stations work fine with it.

argilo:
What gain value does the automatic gain selection choose for that station?

pclov3r:
It will choose the max gain of 49.6 dB.

argilo:
Can you show me screenshots of the FFT with the gain set to 20.7 and 49.6? Also, could you cut & paste the debug messages printed by the gain selection? Thanks.

pclov3r:
Here is a log output

> 16:17:33 INFO  main.c:192: [0] Generic RTL2832U OEM
> 16:17:33 INFO  main.c:192: [1] Generic RTL2832U OEM
> Found Rafael Micro R820T tuner
> Exact sample rate is: 1488375.071248 Hz
> 16:17:34 INFO  main.c:56: Gain: 0.0 dB, Power: -44.875027 dBFS
> 16:17:34 INFO  main.c:56: Gain: 0.9 dB, Power: -42.619064 dBFS
> 16:17:34 INFO  main.c:56: Gain: 1.4 dB, Power: -41.733383 dBFS
> 16:17:34 INFO  main.c:56: Gain: 2.7 dB, Power: -38.573700 dBFS
> 16:17:34 INFO  main.c:56: Gain: 3.7 dB, Power: -36.965065 dBFS
> 16:17:35 INFO  main.c:56: Gain: 7.7 dB, Power: -33.935741 dBFS
> 16:17:35 INFO  main.c:56: Gain: 8.7 dB, Power: -32.194080 dBFS
> 16:17:35 INFO  main.c:56: Gain: 12.5 dB, Power: -29.703524 dBFS
> 16:17:35 INFO  main.c:56: Gain: 14.4 dB, Power: -28.519672 dBFS
[...]

PR #72, opened 2017-09-24
Merge experimental branch

argilo:
The experimental branch has got lots of nice improvements in it, so let's merge it once you have a chance to test it @awesie.

I'd appreciate feedback on the experimental branch from anyone else who can try it.

/cc @classicjazz @pclov3r @docprofsky @KeyserSoze1 @itdaniher @tylert.

PR #82, opened 2017-10-01
Go back to the old gain selection method.

argilo:
It's apparent from @pclov3r's report #81 that the new automatic gain selection method doesn't always work well. This PR reverts the change, but also reduces the number of FFTs performed by the automatic gain selection from 3072 to 256, so it should be just as fast as it was after my change.

@pclov3r How does this work for you?

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -11,6 +11,8 @@
 
 #define FATAL_EXIT(x,...) do { log_fatal(x, ##__VA_ARGS__); exit(1); } while (0)
 
+#define SNR_FFT_COUNT 256
+
 // FFT length in samples
 #define FFT 2048
 // cyclic preflex length in samples
--- a/src/input.c
+++ b/src/input.c
@@ -198,7 +198,7 @@ static void measure_snr(input_t *st, uint8_t *buf, uint32_t len)
         st->snr_cnt++;
     }
 
-    if (st->snr_cnt > 2048)
+    if (st->snr_cnt >= SNR_FFT_COUNT)
     {
         // noise bands are the frequncies near our signal
         float noise_lo = 0;
--- a/src/input.c
+++ b/src/input.c
@@ -181,21 +181,54 @@ void input_wait(input_t *st, int flush)
 #endif
 }
 
-static void measure_power(input_t *st, uint8_t *buf, uint32_t len)
+static void measure_snr(input_t *st, uint8_t *buf, uint32_t len)
 {
-    unsigned int i;
+    unsigned int i, j;
 
-    for (i = 0; i < len; i += 2)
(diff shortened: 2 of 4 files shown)


argilo:
@classicjazz You were also interested in the automatic gain selection, so I'd love to get your feedback on this change too.

argilo:
The change to increase the speed of the old algorithm is in 737edbdbdeaa68aff190789396809937cf35ac1c.

pclov3r:
This seems to be more reliable and appears to clear up the issues I reported in https://github.com/theori-io/nrsc5/issues/81

argilo:
Great, thanks for testing!

pclov3r:
I think the master branch needs a pull request as well for this as https://github.com/theori-io/nrsc5/pull/72 included this.

pclov3r:
Sorry, The master branch needs a pull request for this fix as https://github.com/theori-io/nrsc5/pull/72 included the auto gain function that was reverted and changed here due to being unreliable :)

argilo:
@pclov3r This will make it into the master branch the next time we merge in the experimental branch, which I expect will be soon.

classicjazz:
Total time to evaluate AGC is also 1 second. So, this is equivalent time wise to the current method.

For a strong signal (WHYY):
Time after that: 5 seconds to "Synchronized"
Then another 3 seconds to playing in MPV.

For a moderate signal (WRTI):
Similar AGC time, add an additional second each for Synchronized and audio playback.

For a weak signal (WXPN), 
Time after AGC to synchronized = 19 seconds
No audio is ever heard though I see a lot of these messages

> 10:08:39 INFO  main.c:65: Gain: 49.6 dB, CNR: 2.012946 dB
10:08:58 INFO  sync.c:248: Synchronized!
10:08:59 INFO  sync.c:324: MER: -12.290449 dB (lower), -13.312159 dB (upper)
10:08:59 INFO  decode.c:77: BER: 0.161512, avg: 0.161512, min: 0.161512, max: 0.161512
10:09:25 INFO  acquire.c:104: Timing offset: 434.750000, slope: 1.015625 (adjust)
10:09:32 INFO  acquire.c:104: Timing offset: 412.062500, slope: -1.378906 (adjust)
10:09:35 INFO  acquire.c:104: Timing offset: 416.875000, slope: 1.132812 (adjust)

I have been testing a number of different antennas--amplified and not--to determine the best reception. I am currently using a Yagi antenna with amplifier. Which begs the question, would users be better off going all digital amplification, using NSRC alone?

classicjazz:
Just some additional info: assuming Wikipedia is correct, then WHYY, WRTI and WXPN all broadcast from the same antenna farm about 15 miles away. The difference is that WHYY is 13500 watts ERP, WRTI is 7700 ERP, and WXPN is 2650 watts ERP. Further, WXPN is only 105 watts digital.

Issue #68, opened 2017-09-22
Pop / click suppression


classicjazz:
I have noticed that when a signal is low, there may be audible pops / clicks / or 1 second "whistles" in the feed from NRSC to MPV.

This seems to coincide with these log notices:
WARN  frame.c:511: crc mismatch!

Is there a way that we could suppress the pops with either 
1) worst case, silence or
2) preferably, an interpolation of the missing data (analog wave)

I have unsuccessfully searched for FFMPEG / MPV filters similar to Audacity's pop/click removal to solve this in real time on the audio stream. 

But it really would be better to have this functionality in NRSC because it can be correlated with the CRC mismatches.

pclov3r:
Seems like this is still kinda an issue? This however has been open for quite some time.

argilo:
Closing, as this should have been resolved by #399.

PR #83, opened 2017-10-01
Prevent audio stutter by adding silence to start of live audio

argilo:
Audio stutter often occurs at the start of live audio playback, because NRSC-5's audio codec has a variable bit rate, but nrsc5 doesn't account for that.

To solve the problem, I've added 10 packets worth of silence to the beginning of the audio. I chose 10 because it's the maximum allowed elastic buffer depth. See http://www.nrscstandards.org/sg/nrsc-5-d-reference-docs/1017s.pdf section 4.3 for details.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -169,6 +169,7 @@ void decode_reset(decode_t *st)
     st->ready_p3 = 0;
     memset(st->pt_p3, 0, sizeof(unsigned int) * 4);
     pids_init(&st->pids);
+    output_begin(st->input->output);
 }
 
 void decode_init(decode_t *st, struct input_t *input)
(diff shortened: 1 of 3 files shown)


awesie:
I assume this is only an issue when we are decoding HDC ourselves, since other media players usually do their own buffering?

argilo:
Probably so.

awesie:
It looks like it fixes the issue, but I think a better fix would be an internal buffer of 10 packets. I'm fine with committing this now so that it can go into master with #82.

argilo:
I've noticed that stutter can still occur if the signal is weak enough that CRC mismatches occur, but not so weak that synchronization is lost and reacquired. I guess a CRC mismatch should cause silence to be played.

awesie:
Probably. That issue has some overlap with #68.

PR #84, opened 2017-10-01
Search for symbol synchronization over the whole period.

argilo:
OFDM timing synchronization fails about 5% of the time, because it only searches for the start of a symbol over the range 0-2047, even though it could occur at any sample offset from 0 to 2159. Running the following command demonstrates the problem:
```
xz -d < ../support/sample.xz | tail -c +561 | src/nrsc5 -r - 0
```
To fix this, I've expanded the search window to include the entire range (i.e. the duration of one symbol, including cyclic prefix).


awesie:
I think this is good. I had to convince myself that it didn't read past the end of buffer.

PR #90, opened 2017-10-17
Refactor frame_process.

argilo:
There was a lot of duplication between `frame_process` and `find_program`, so I decided to merge them. This will also open up the possibility of having `frame_process` deal with multiple programs, for instance so that multiple programs can be saved to disk or streamed simultaneously.

`frame_process` now parses the frame in a single pass from beginning to end. While I was at it, I made some other improvements:

* proper HEF parsing (in `parse_hef`)
* additional length checks for robustness
* simplified `parse_location` by passing in the start of the location bytes.

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -26,6 +26,7 @@
 
 #define BBM 0x42E23A7D
 #define MAX_AAS_LEN 8212
+#define MAX_AUDIO_PACKETS 64
 
 typedef struct
 {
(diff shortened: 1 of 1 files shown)


argilo:
I tested this with recordings from 15 stations, and everything seems to work well.

argilo:
It may be easier to review the changes with whitespace excluded: https://github.com/theori-io/nrsc5/pull/90/files?w=1

awesie:
I don't see any obvious issues. I definitely think it is an improvement over the old code.

argilo:
Thanks for reviewing @awesie!

PR #91, opened 2017-10-18
Process all programs in frame_process.

argilo:
I've changed `frame_process` to process all programs, and moved the decision about which program to play into the output component. That will make it easier to output multiple programs in the future.


argilo:
The changes in `frame.c` are much easier to review with whitespace changes ignored: https://github.com/theori-io/nrsc5/pull/91/files?w=1

PR #94, opened 2017-10-21
Add more detail to usage examples.

### LINKED ISSUE #93: Update README to show example of usage without piping to external player
In light of https://github.com/theori-io/nrsc5/issues/92 I discovered that the README under example usage doesn't include an example of how to use it without piping the audio to a external player. 

Probably would be helpful for new users if there was a example such as `nrsc5 90.5 0` to indicate that it doesn't need to be piped to a external player.

argilo:
Fixes #93.

I've updated the usage examples to begin with the simplest possible invocation (which should be suitable for most users), and added explanations for each of the examples.

How does this look @pclov3r @impala454?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ You can test the program using the included sample capture:
 
 This was designed for use with an RTL-SDR dongle since that was our testing platform.
 
-Options:
+### Options:
 
        frequency                       rtl-sdr center frequency in MHz or Hz
                                          (do not provide frequency when reading from file)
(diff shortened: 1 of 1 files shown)


pclov3r:
Looks good to me!

README.md
```diff
@@ -59,10 +59,20 @@ Options:
                                          (1 = DEBUG, 2 = INFO, 3 = WARN)
        -v                              print the version number and exit
 
-Examples:
+### Examples:
 
-     $ nrsc5 -p 63 -g 490 -w samples1071 107.1 0
```
argilo: I removed the `-p` option from this example because there's no need to manually set PPM correction now that nrsc5 corrects the frequency offset automatically.

impala454:
Looks good!  Thanks for the clarification.

PR #96, opened 2017-10-25
Play audio packets immediately; don't wait for number 0.

argilo:
Currently `frame_process` waits for an audio packet with sequence number 0 to come along before it starts passing audio packets along for decoding. This can add up to a couple seconds to the start-up time.

As far as I know, the sequence numbers are simply indexes into a 64-frame circular buffer that a player could use. There's nothing special about sequence number 0, so I don't see any reason to wait for it to arrive. Everything seems to work fine after this change.

When combined with #95 this brings the acquisition time down pretty close to the theoretical minimum.

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -461,7 +461,7 @@ void frame_process(frame_t *st, size_t length)
     while (offset < length - 96)
     {
         unsigned int start = offset;
-        unsigned int j, seq, lc_bits, loc_bytes, prog;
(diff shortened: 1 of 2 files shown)


argilo:
@classicjazz @pclov3r This change takes care of the other significant source of start-up delay. It should make about a 1-second improvement on average.

argilo:
If you want to try the two improvements together, just check out the `fast-start` branch: https://github.com/theori-io/nrsc5/tree/fast-start

PR #101, opened 2017-10-29
Fix MER calculation for primary mode 3.

argilo:
I noticed that the reported MER is 0.8 dB low for stations using primary mode 3, because the signal is assumed to come from only 10 partitions, whereas the noise is summed from all 12. I corrected this simply by changing `P1_DATA_PER_BAND` to `partitions_per_band * 18`.

Since `P1_DATA_PER_BAND` was no longer used, I removed it from defines.h, and I also got rid of `P1_BAND_LENGTH` (which was barely used).

Finally, I changed the MER display to include only one decimal place. (Previously it showed the default six.)

Here's an example for a recording of my local station. Before:
```
08:31:27 INFO  sync.c:315: MER: 10.645880 dB (lower), 10.632761 dB (upper)
08:31:28 INFO  sync.c:315: MER: 10.610696 dB (lower), 10.569898 dB (upper)
08:31:28 INFO  sync.c:315: MER: 10.646533 dB (lower), 10.668467 dB (upper)
08:31:29 INFO  sync.c:315: MER: 10.632702 dB (lower), 10.736723 dB (upper)
08:31:31 INFO  sync.c:315: MER: 10.763515 dB (lower), 10.707469 dB (upper)
```
After:
```
08:31:43 INFO  sync.c:315: MER: 11.4 dB (lower), 11.4 dB (upper)
08:31:44 INFO  sync.c:315: MER: 11.4 dB (lower), 11.4 dB (upper)
08:31:44 INFO  sync.c:315: MER: 11.4 dB (lower), 11.5 dB (upper)
08:31:45 INFO  sync.c:315: MER: 11.4 dB (lower), 11.5 dB (upper)
08:31:47 INFO  sync.c:315: MER: 11.6 dB (lower), 11.5 dB (upper)
```
The values are unchanged for stations using primary mode 1.

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -20,10 +20,6 @@
 #define FFTCP (FFT + CP)
 // OFDM symbols per L1 block
 #define BLKSZ 32
-// number of data subcarriers per sideband
-#define P1_DATA_PER_BAND 180
-// number of subcarriers per sideband
-#define P1_BAND_LENGTH (11 + P1_DATA_PER_BAND)
 // index of first lower sideband subcarrier
 #define LB_START (1024 - 546)
 // index of last upper sideband subcarrier
--- a/src/defines.h
+++ b/src/defines.h
@@ -20,6 +20,8 @@
 #define FFTCP (FFT + CP)
 // OFDM symbols per L1 block
 #define BLKSZ 32
+// number of primary main partitions
+#define PM_PARTITIONS 10
 // index of first lower sideband subcarrier
 #define LB_START (1024 - 546)
 // index of last upper sideband subcarrier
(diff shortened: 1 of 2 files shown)


awesie:
In general, I was trying to avoid any magic numbers, hence P1_BAND_LENGTH instead of 180. I agree that it is probably a lost cause at this point.

argilo:
I was thinking about that too, and I thought maybe it would make more sense to have `P1_PARTITIONS` be 10 and then to use `P1_PARTITIONS * 19` in the 10 or so places where we have 190 currently.

argilo:
@awesie I added a second commit with that change. I used the name `PM_PARTITIONS` because it's really the primary main partitions; in primary modes 5 and 6 things other than P1 go there.

PR #105, opened 2017-11-07
Improve frequency tracking.

argilo:
Cheaper RTL-SDR dongles that lack a TCXO can drift quite a bit (~10 ppm) in frequency as they warm up. Currently nrsc5 doesn't account for this; it expects the frequency offset to remain nearly constant.

To fix this, I've changed the frequency tracking code to calculate the difference between the angles of `max_v` and `cexpf(I * st->prev_angle)` and adjust the angle using that value. That allows the angle to wrap around the full circle multiple times if necessary to track the drift of the RTL-SDR's oscillator.

Also I changed the smoothing factor from 0.5 to 0.1 as this made a significant improvement to the BER of the 15 recordings I use for tests.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -26,7 +26,7 @@
 void acquire_process(acquire_t *st)
 {
     float complex max_v = 0;
-    float angle, max_mag = -1.0f;
+    float angle, angle_diff, angle_factor, max_mag = -1.0f;
     unsigned int samperr = 0, i, j, keep;
     unsigned int mink = 0, maxk = FFTCP;
 
@@ -63,16 +63,9 @@ void acquire_process(acquire_t *st)
         }
     }
 
-    // limited to (-pi, pi)
-    angle = cargf(max_v);
-    if (st->prev_angle)
-    {
-        if (st->prev_angle > M_PI*15/16 && angle < -M_PI*15/16)
-            angle += M_PI * 2;
-        else if (st->prev_angle < -M_PI*15/16 && angle > M_PI*15/16)
-            angle -= M_PI * 2;
-        angle = 0.5 * st->prev_angle + 0.5 * angle;
-    }
+    angle_diff = cargf(max_v * cexpf(I * -st->prev_angle));
+    angle_factor = (st->prev_angle) ? 0.1 : 1.0;
+    angle = st->prev_angle + (angle_diff * angle_factor);
     st->prev_angle = angle;
 
     for (i = 0; i < M; ++i)
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -64,7 +64,7 @@ void acquire_process(acquire_t *st)
     }
 
     angle_diff = cargf(max_v * cexpf(I * -st->prev_angle));
-    angle_factor = (st->prev_angle) ? 0.1 : 1.0;
+    angle_factor = (st->prev_angle) ? 0.25 : 1.0;
     angle = st->prev_angle + (angle_diff * angle_factor);
     st->prev_angle = angle;
 


pclov3r:
I upgraded all my RTL-SDRs to ones that use TCXO LO's

I'll give this a test with a cheap ones without it and see how this change helps.

argilo:
It turns out the frequency of non-TCXO dongles can drift fairly quickly at times, and so I had to make the averaging a bit less aggressive. I changed the smoothing factor to 0.25 and that improved the BER on non-TCXO dongles during warmup at the expense of a small decrease in BER for TCXO dongles.

argilo:
@pclov3r Prior to this change, you should find that if you plug in a non-TCXO dongle and start receiving, it will lose sync within 1-2 minutes because the frequency tracking will fail as the oscillator warms up. After the change everything should work fine with both TCXO and non-TCXO dongles.

awesie:
I'm overseas, so not able to test it. Was this issue previously hidden by the resampler which was removed in #95?

pclov3r:
Would it be too much work to include a branch that has the improvements for people using TCXO dongles so they can have the BER improvements?

argilo:
@awesie I don't expect the resampler would have helped, as it was only able to correct for error in the sample clock, not the LO.

pclov3r:
@argilo To follow up my comment how much of a BER difference is there on TCXO dongles with and without this change?

If it's extremely minor I guess it wouldn't make sense to have a separate branch for that.

argilo:
> Would it be too much work to include a branch that has the improvements for people using TCXO dongles so they can have the BER improvements?

@pclov3r Currently nrsc5 relies entirely on pre-FFT synchronization, which is not very accurate. Averaging helps reduce the error, but the real fix is to switch to post-FFT synchronization once the signal is acquired. I'm working on this, and it improves the BER more than any amount of averaging here would.

Thus, I don't think it's worth making the averaging configurable; post-FFT synchronization should make things better for both types of dongles.

pclov3r:
@argilo Ah, That does sound like a better solution all around.

argilo:
Also, I should add that most of the BER improvement came from changing the averaging factor from 0.5 to 0.25; moving to 0.1 made things only slightly better for TXCO dongles but significantly worse for non-TCXO dongles.

argilo:
@waesie To confirm, I just tested the code prior to the removal of the resampler, and it fails the same way with a non-TCXO dongle. Receive failed after 34 seconds.

argilo:
@awesie I just hadn't noticed the problem until recently as I was always testing with a dongle that had a TCXO.

awesie:
Thanks. I was just curious because I don't think my dongle has a TXCO but I don't remember it failing that quickly.

I would go ahead and merge since it appears to be a strict improvement.

argilo:
It could be that your dongle was already warmed up when you started receiving. You'll see the worst behaviour if you plug in the dongle and immediately fire up nrsc5.

Another possibility is that your frequency offset happened to put the angle close to -pi with a cold dongle, and when it warmed up it didn't exceed +pi. The problem manifests whenever the angle increases above +17/16*pi or below -17/16*pi. The new code allows the angle to increase or decrease indefinitely.

classicfm:
This is interesting, as I don't have a TXCO radio either. I've come close to replacing ("upgrading") it with an RTL-SDR.com dongle a few times, but never have. I really have not had a problem with it.

I have not seen mine lose sync like it seems it should have, but reading the comments here it seems as though if it remains plugged in all the time this issue doesn't apply? My main use for the dongle has been recording classical music radio shows from FM radio. I do this via cron, so needless to say it's always plugged in. 

It would be great if my radio station was broadcasting in HD, but either they're not, or they're operating under the old 1% dBc limit, which would be a 185 watt ERP signal at 13 miles. If this situation changes, I'd use nrsc5 for at least three hours of recording every day and possibly more, which would be a great test case.

argilo:
@classicfm There is still some temperature change (and resulting frequency drift) when you go from idle to receiving, but probably not as much as when you plug in a cold RTL-SDR and immediately start receiving. Also, it's likely that the amount of drift varies between models. So you may also have been fortunate not to hit this bug.

PR #106, opened 2017-11-13
Implement post-FFT synchronization.

argilo:
Most OFDM receivers use pre-FFT synchronization only during acquisition, then switch to post-FFT synchronization. This PR adds post-FFT synchronization to nrsc5. The change improves the bit error rate across all recordings I tested with.

Since post-FFT synchronization requires feedback to pass from `sync.c` back to `acquire.c` (preferably immediately) I decided to remove the input and sync threads. I believe they were only added to make reception possible on multi-core ARM processors, but performance has since been improved in other ways. (The output thread is still useful so that signal processing can continue even when audio output is blocked, so I left it in place.)

I also managed a significant speedup (especially on ARM) by replacing the complex exponentiation in `acquire_process` with a multiplication. CPU utilization on a Raspberry Pi 3 is now around 55%, compared to 160% previously.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -25,9 +25,10 @@
 
 void acquire_process(acquire_t *st)
 {
-    float complex max_v = 0;
+    float complex max_v = 0, phase_increment;
     float angle, angle_diff, angle_factor, max_mag = -1.0f;
-    unsigned int samperr = 0, i, j, keep;
+    int samperr = 0;
+    unsigned int i, j, keep;
     unsigned int mink = 0, maxk = FFTCP;
 
     if (st->idx != FFTCP * (M + 1))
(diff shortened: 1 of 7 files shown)


argilo:
@pclov3r @classicjazz @classicfm @tylert I'd be interested to get your feedback on this change. It should improve reception of weaker signals.

pclov3r:
@argilo great work as always. 

For me it seems to improve BER as well and perform better with weak signals.

src/sync.c
```diff
@@ -276,11 +287,25 @@ void sync_process(sync_t *st, float complex *buffer)
...
         for (i = 0; i < partitions_per_band * 19; i += 19)
         {
             adjust_data(buffer, st->phases, LB_START + i, LB_START + i + 19);
             adjust_data(buffer, st->phases, UB_END - i - 19, UB_END - i);
+
+            samperr += phase_diff(st->phases[(LB_START + i) * BLKSZ], st->phases[(LB_START + i + 19) * BLKSZ]);
+            samperr += phase_diff(st->phases[(UB_END - i - 19) * BLKSZ], st->phases[(UB_END - i) * BLKSZ]);
+        }
+        samperr = samperr / (partitions_per_band * 2) * 2048 / 19 / (2 * M_PI);
+        st->samperr = roundf(samperr);
```
argilo: Here's the meat of the change. Symbol timing error causes the subcarrier phases to increase linearly with carrier number. By computing the average phase increase between neighbouring reference subcarriers, we get a very good estimate of the timing err (`samperr`).

src/sync.c
```diff
@@ -276,11 +287,25 @@ void sync_process(sync_t *st, float complex *buffer)
...
+
+        for (i = 0; i < partitions_per_band * 19 + 1; i += 19)
+        {
+            angle += prev_slope[LB_START + i];
+            angle += prev_slope[UB_END - i];
         }
+        angle /= (partitions_per_band + 1) * 2;
+        st->angle = angle;
```
argilo: Frequency error causes a linear increase in the phase of each carrier over time, which we're already measuring as `slope`. The average slope gives a very good estimate of the frequency error.

src/acquire.c
```diff
@@ -25,72 +25,87 @@
...
 
-    memset(st->sums, 0, sizeof(float complex) * FFTCP);
-    for (i = mink; i < maxk + CP; ++i)
-    {
-        for (j = 0; j < M; ++j)
-            st->sums[i] += st->buffer[i + j * FFTCP] * conjf(st->buffer[i + j * FFTCP + FFT]);
+        angle_diff = -st->input->sync.angle;
+        st->input->sync.angle = 0;
+        angle = st->prev_angle + angle_diff;
+        st->prev_angle = angle;
```
argilo: The post-FFT timing & frequency error estimates are used once synchronization is achieved, thereby avoiding the need to compute the (less accurate) pre-FFT estimates.

classicfm:
I do not see any practical change in my installation. WCNY (classic FM) is a lost cause, the COFDM sidebands are nonexistent when looking in GQRX. I emailed the station by the contact form on the Web and have not received a response. I may call them at some point. I have also considered taking my SDR and laptop down closer to their transmitter to see if there's a difference, but from everything I've read and based on other stations I can receive, I find it extremely unlikely that the hardware or software on my end is to blame.

That said, there's another test case. An old-style class A (5.7 kW) at 12.6 miles. FMFool says it is not a line of sight reception for me (it is "2Edge", I forget what that means). I can get a lock, both with these changes and without, though I have never and still don't get to an audio stream. However, BER drops from 0.22 average to 0.18 with these changes.

I am sure I could move my antenna and then get the stream without issue, but the transmitter is nearly 180° opposite the stations I care about (and in fact is opposite most other stations, TV and FM, atop the hills down south).

classicfm:
I do have to ask, however: On occasion, I will get what appear to be spurious "First block @ N" messages on a station that is not broadcasting in HD, WCNY included. This began once the resampler was removed, though prior to that the message was "Timing offset / slope" in a similar situation. I have seen this occur with a station that does not broadcast in HD, according to HDRadio.com and Arbitron. I looked at the source code and, though I am not familiar enough with things to know for sure, it appears to me that these messages _could_ result from noise. 

I have not seen these with the post-FFT branch, but they were rare before, so I have not had enough of a chance to say they are gone.

src/sync.c
```diff
@@ -202,9 +210,12 @@ void sync_process(sync_t *st, float complex *buffer)
 
     for (i = 0; i < partitions_per_band * 19 + 1; i += 19)
     {
+        prev_slope[LB_START + i] += st->angle_adj;
```
awesie: These two lines have an unintended side-effect during startup due to line 53. Line 53 is a hack that uses prev_slope[...] = 0 has a flag that we don't have any previous slope information.

argilo:
> Is this assessment correct? If I see "First block @ N", that does not necessarily mean that there is an HD signal, right?

Correct. The current detection algorithm in `sync_process` is prone to false positives. It's on my list of things to fix.

argilo:
The reason is that it only checks a handful of bits in two particular reference subcarriers. It's easy for those bits to have the right values by chance when there is no signal. And if there's a strong signal but one of the two subcarriers is blocked by a noise spike, detection would fail completely. I think the proper way to do it is to check all reference subcarriers, and accept if some threshold (e.g. half) are valid.

src/sync.c
argilo: I'll check into that tonight, but I suspect there will be phase information by the time this line is reached. It could be there's a better way to implement it though.

awesie: If you look at my gist I linked, my solution was to fix the hack at line 53. These lines should be fine, but their addition made the hack problematic.

argilo: > by the time this line is reached

I meant to say: by the time `st->angle_adj` is non-zero.

argilo:
Thanks for the comments @awesie. I'll make some improvements tonight.

awesie:
For your reference, I uploaded the particular test file that I was fixing to http://server1.awesie.org/1071.37.2.

argilo:
@awesie Your changes looked good except for one bug which I fixed in 4b73a859b717260321662e9d3ee67ea4b678b136. The `angle_adj` value is only meant to adjust the previous slope, not the current one. That fix improved the BER of your recording, as well as most of mine.

PR #107, opened 2017-11-13
Speed up adjust_data and remove fast_cexpf.

argilo:
This change builds on top of #106 and gives another significant speedup simply by moving the expensive complex exponentiations out of the innermost loop of `adjust_data`.

The speedup is around 12% on Intel, and on my Raspberry Pi 3 the CPU utilization is reduced to around 50%.

Since complex exponentiations are no longer a bottleneck, I removed `fast_cexpf`.


argilo:
Bit error rates are the same before & after this change, so I expect it's correct.

pclov3r:
Looks like this PR is removing the unsafe math optimizations?

If so perhaps that should be removed from the readme as well.

argilo:
@pclov3r I haven't completely removed `USE_FAST_MATH` yet; it still reduces the size of the FIR filter. I'll remove that in a future PR.

awesie:
This looks good to me. Feel free to merge with #106 once it is ready.

PR #108, opened 2017-11-14
Merge experimental branch.

argilo:
Since the last merge, there have been some significant performance and sensitivity improvements, as well as a bug fix. It's probably worth merging to master, as long as things are working well. So far things look good.

Feedback appreciated @pclov3r @classicjazz @classicfm @tylert.

PR #109, opened 2017-11-16
Compensate for sample clock error in post-FFT sync.

argilo:
This fixes a regression observed by @awesie in https://github.com/theori-io/nrsc5/pull/108#issuecomment-344143640

The new post-FFT synchronization code was performing sub-optimally for non-TCXO dongles, because it assumed the sample clock error to be zero. To correct this, I've added code to estimate the sample clock error and add a corresponding amount to the symbol timing error (`samperr`) to compensate. This accounts for the number of samples that will be gained or lost between invocations of `acquire_process`.

The sample clock error is estimated by performing a linear regression on the `prev_slope` values. The slope gives the sample clock error, while the intercept (already computed as `angle`) gives the frequency error.

After this change, the bit error rate either improved (in some cases, significantly) or stayed approximately the same for all of the 40 recordings I now test with.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -20,9 +20,6 @@
 #include "defines.h"
 #include "input.h"
 
-#define SYMBOLS 2
-#define M (BLKSZ * SYMBOLS)
-
 void acquire_process(acquire_t *st)
 {
     float complex max_v = 0, phase_increment;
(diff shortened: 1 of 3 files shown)


awesie:
Great work as always.

Issue #80, opened 2017-09-30
Analog to Digital migration behavior to match a psychical HD radio?


pclov3r:
Now, I'm thinking this may be far too much work with little benefit but I'll throw the idea out anyway.

Is it possible to start with Analog FM and than once the digital channel has been acquired than switch to digital? 

Basically the goal of matching the behavior of how a physical HD radio works. 

Thanks!

argilo:
It probably would be a lot of work to add that feature. It's not something I'm particularly interested in, but I'll leave the issue open in case someone else feels like implementing it someday.

awesie:
The difficult part is figuring out how much delay we have processing the digital audio. And then combine this with the information in the PDU header to determine how much we need to delay the analog audio.

classicjazz:
Unlike a moving car, NSRC5 is likely played on a stationary computer. So, my  preference would be pure digital reception. But speeding up initial audio playback is desirable.

There's been a lot of progress made on AGC timing (reducing it from 3-5 seconds to 1 second). 

It seems like the next things to focus on are reducing the time to "Synchronized" (5-19 seconds) and then audio playback (3-4 seconds). See more anecdotal details here: https://github.com/theori-io/nrsc5/pull/82#issuecomment-333350394

argilo:
@classicjazz There is a certain amount of latency inherent in the NRSC-5 standard (primarily coming from the interleavers) and so the acquisition time will always be at least a few seconds. But I believe there is still quite a bit of room for improvement in nrsc5.

pclov3r:
@awesie @argilo I agree. It seems the focus should be put on getting it to acquire and sync faster if possible vs adding analog FM and waiting for it to sync and than switching.

classicjazz:
@argilo 

I timed tuning secondary channels on the HD radio in my car. It consistently takes 5 seconds to jump between stations' HD2 channels from the time it is selected until audio is played. This seems like a useful benchmark for the art of the possible.

classicfm:
How about an option to exit with an error if not synchronized after a certain time? This would allow applications using nrsc5 to failover to using RTL_FM or SoftFM.

I read somewhere that many stations don't have HD capability on their back up equipment, for cost reasons. These backup facilities get used more than you think, for maintenance or whatever, and so we have all become used to 24/7 FM always being there. With the migration behavior of physical radios, I'd bet most people don't even notice when HD is down.

pclov3r:
@classicfm that could work.

[...]

PR #114, opened 2017-11-21
Switch to a halfband filter

argilo:
Now that post-FFT synchronization is place, the expensive (32-tap) band-pass filter is no longer required, except during the brief time when the signal is initially acquired using pre-FFT synchronization. The band-pass filter can be replaced by a halfband filter, which only requires as much computation as a 4-tap filter.

This PR makes the following changes:
* remove `USE_FAST_MATH`, since the halfband filter is already very fast
* change the 32-tap band-pass filter in input.c to a 4-tap halfband filter (equivalent to a 17-tap low-pass FIR filter)
* change input.c to pass `cint16_t` samples to acquire.c (instead of `float complex`)
* move CFO correction from input.c to acquire.c
* decrease the gain of `U8_Q15` by 2 (since the halfband filter has a gain of 2)
* add a band-pass filter to acquire.c, for use in pre-FFT synchronization (used during initial acquisition)
* narrow the passband of the band-pass filter (to better eliminate signals outside the OFDM sidebands)
* change the band-pass filter to 31 taps (so that the delay is an even 15 samples instead of 15.5)
* fold the non-NEON dot product calculation (to reduce the number of multiplications by a factor of 2)

These changes reduce the amount of computation by about 35% on Intel and 13% on ARM, and slightly improve the BER across my suite of 40 test signals.

Another advantage of switching to a halfband filter is that it doesn't filter out the analog FM signal (which would allow blending from analog to digital as suggested in #80). It would also allow all-digital signals to be decoded.

Testing would be appreciated @classicfm @classicjazz @pclov3r @tylert.


argilo:
The frequency responses of the new filters are as follows:

halfband

bandpass

pclov3r:
Great work!

Seems to make the same improvements here that you are seeing.

I'm curious about the band-pass filter narrowing and if that may cause changes for non-tcxo dongles. Perhaps https://github.com/theori-io/nrsc5/pull/109 solves that issue.

argilo:
@pclov3r The band-pass filter should have little effect on receive performance, because its output is *only* used to compute pre-FFT symbol timing & frequency estimates; the rest of the demodulation process uses the halfband-filtered signal (even during acquisition).

But your comment did make me wonder whether those pre-FFT estimates are better than they were previously, so I checked the BER with post-FFT synchronization disabled. (This can easily be accomplished by changing `if (st->input->sync.ready)` to `if (0)` in acquire.c.) In almost all cases, the new design improves the BER; overall it's reduced by about 19%. That convinces me that the narrower band-pass filter is an improvement.

tylert:
@argilo it builds and works great for me on Debian Stretch x86_64 with SSE=on (Nooelec dongle with TCXO).

argilo:
Thanks for testing!

classicjazz:
Working with the experimental branch, post merge:

I can tune all my test stations (an improvement!) but still having issues staying synch'ed without CRC errors with WHYY.

If I try to create a dump file using 
nrsc5 -o out.adts -f adts $FREQUENCY $BAND

I get 
18:07:49 INFO  main.c:185: [0] Generic RTL2832U OEM
nrsc5(68007,0x7fffb4745340) malloc: *** error for object 0x7ffee9bd8928: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
/Users/mike/Documents/Scripts/HD Radio Dump.sh: line 34: 68007 Abort trap: 6           nrsc5 -o out.adts -f adts $FREQUENCY $BAND

every single time though it changes somewhat...

Tuning to the radio station: 90.9 Mhz HD1
18:10:07 INFO  main.c:185: [0] Generic RTL2832U OEM
nrsc5(68015,0x7fffb4745340) malloc: *** error for object 0x7ffee335d928: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
/Users/mike/Documents/Scripts/HD Radio Dump.sh: line 34: 68015 Abort trap: 6           nrsc5 -o out.adts -f adts $FREQUENCY $BAND

I am currently capturing an hdc dump for WHYY for your testing purposes.

argilo:
@classicjazz A capture of the RF samples (made with the `-w` option) would be much more useful.

classicfm:
@classicjazz those numbers are memory addresses, they change because the program is never loaded into the same memory location twice. If you haven't already, delete your source tree, clone anew from git, and try again. It *could* be stale object files mixing with new ones.

@argilo Sorry I can't test very well, I'm out of town and distracted. I did run it by SSH, for WAER (which has always been strong and clear), it seems okay. I'll be more useful when WCNY gets their new antenna!

Issue #115, opened 2017-11-24
Images are being received from all streams


cmnybo:
When listening to a station with more than one stream, the album art images are being received from all streams.
Would it be possible to only receive images from the currently playing stream or indicate which stream the image belongs to?

I am trying to write a GUI for nrsc5, and currently there is no way to tell which image belongs to what stream, so I can't show the correct one.

classicfm:
I would like to see the program associated data currently printed as debug log messages instead printed as JSON for this kind of application. Perhaps when an AAS file is received the stream it is associated with could be noted in the json stream.

argilo:
@cmnybo Album art is mostly undocumented, so it will take some reverse engineering to figure out how to properly process the images. This article mentions that the XHDR ID3 tag is used for image synchronization: http://www.radioworld.com/digital-radio/0014/artist-experience-and-album-art-via-hd-radio/330814

I added some code in #116 to dump out XHDR tags, which may help with reverse engineering. It could also be that the port number determines which program the image belongs to.

I'm excited to see what you come up with for a GUI!

argilo:
@cmnybo It looks like the association between programs (HD1, HD2, HD3) and ports is laid out in the "port info" message, which is parsed by the `parse_port_info` function:

https://github.com/theori-io/nrsc5/blob/02cd65b8e4e96a7a573b06364fbb236a915640a8/src/output.c#L404-L462

For my local station, it produces the following:
```
19:34:03 DEBUG output.c:483: 40 01 00 37
19:34:03 DEBUG output.c:497: Found MAJIC 100.3
19:34:03 DEBUG output.c:506: Port 800, type 3, size 265
19:34:03 DEBUG output.c:506: Port 801, type 3, size 265
19:34:03 DEBUG output.c:483: 40 02 00 F2
19:34:03 DEBUG output.c:497: Found 580 CFRA on 100.3 HD2
19:34:03 DEBUG output.c:506: Port 802, type 3, size 265
19:34:03 DEBUG output.c:506: Port 803, type 3, size 265
19:34:03 DEBUG output.c:483: 40 03 00 46
19:34:03 DEBUG output.c:497: Found TSN 1200 on HD3
19:34:03 DEBUG output.c:506: Port 804, type 3, size 265
19:34:03 DEBUG output.c:506: Port 805, type 3, size 265
19:34:03 DEBUG output.c:483: 41 0A 00 18
19:34:03 DEBUG output.c:497: Found HD TMC
[...]

PR #116, opened 2017-11-25
Decode more ID3 tag types.

argilo:
Currently, only TIT2 (title), TPE1 (artist) and TALB (album) ID3 tags are decoded. This PR adds decoding for three additional tags:

* TCON (genre)
* UFID (unique file identifier)
* COMR (commercial)

It also prints out the raw bytes of any other tags. In my set of 40 recordings, only the XHDR tag cannot be decoded. XHDR is used for album art synchronization (according to this article: http://www.radioworld.com/digital-radio/0014/artist-experience-and-album-art-via-hd-radio/330814), and so reverse engineering it may help with #115.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -396,6 +396,72 @@ static void output_id3(uint8_t *buf, unsigned int len)
             log_debug("Album: %s", album);
             free(album);
         }
+        else if (memcmp(buf + off, "TCON", 4) == 0)
+        {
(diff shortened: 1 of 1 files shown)


pclov3r:
Looks like the XHDR tag works here. 

`14:11:09 DEBUG output.c:463: XHDR tag: 36 75 4B BE 01 00`

PR #117, opened 2017-11-27
Use a Costas loop for synchronization.

argilo:
Currently nrsc5 uses an ad-hoc method for BPSK synchronization. Switching to a Costas loop improves the BER for all recordings I tested with. On average, the improvement is 21%.

Also, I added an extra `cfo` parameter to `adjust_ref` to compensate for the rotation introduced by the carrier frequency offset. This simplifies the Costas loop implementation, and removes the need to invert the data in `find_ref`.

The Costas loop processes symbols one at a time (rather than in blocks of 32), which should allow for some future simplifications (such as removing `input_set_skip`).

Testing would be appreciated @classicfm @classicjazz @tylert @pclov3r.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -120,8 +120,10 @@ void acquire_process(acquire_t *st)
     for (i = 0; i < FFTCP * (ACQUIRE_SYMBOLS + 1); i++)
         st->buffer[i] = cq15_to_cf(st->in_buffer[i]);
 
-    sync_adjust(&st->input->sync, angle_diff);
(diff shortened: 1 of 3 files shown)


argilo:
The design is modelled after GNU Radio's Costas loop: https://github.com/gnuradio/gnuradio/blob/master/gr-digital/lib/costas_loop_cc_impl.cc

pclov3r:
More awesome work.

Appears to work fine here and improvements as you stated.

PR #119, opened 2017-11-27
Only write album art & station logo for the current program.

argilo:
Hopefully this solves #115.

This change does a few things:

* log the port number for files
* store the service data type and program number for each port
* for files with service data type 0x40, only write when the program number matches the current one

Traffic & weather files will still be written out regardless of the current program number.

This code is based on my reverse engineering assumptions from #115. Testing would be appreciated @cmnybo.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -405,6 +405,7 @@ static void parse_port_info(output_t *st, uint8_t *buf, unsigned int len)
 {
     static int dump = 1;
     unsigned int idx = 0;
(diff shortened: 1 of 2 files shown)


cmnybo:
I tested this branch and it is working correctly for me.
The image files from other audio streams are no longer written and the port number is displayed.
The weather and traffic data is still written.

argilo:
Thanks for testing @cmnybo. Hopefully this helps with your GUI work.

PR #121, opened 2017-11-30
Clean up sync code

argilo:
This is just some code cleanup. I've removed a bunch of unused stuff (`BUFS`, `dump_ref`, `ref_buf`, `buf_idx`, `used`) and switched `buffer` and `phases` to two-dimensional array syntax.

BER values are the same before & after the change for all my test files, so it appears this change hasn't broken anything.

### DIFF
--- a/src/sync.c
+++ b/src/sync.c
@@ -21,17 +21,7 @@
 #include "input.h"
 #include "sync.h"
 
-#define BUFS 4
-
-static void dump_ref(uint8_t *ref_buf)
-{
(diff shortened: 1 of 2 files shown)


awesie:
Looks okay to me. I admit I had to lookup the C pointer-to-array, since I don't think I've seen that before.

argilo:
> I admit I had to lookup the C pointer-to-array, since I don't think I've seen that before.

It was new to me too. Found it on Stack Overflow yesterday. :-)

PR #132, opened 2018-04-14
Fix station location parsing

supersat:
This change fixes parsing of the station location. The old code treated the coordinates as an unsigned integer with a sign bit. However, coordinates are actually sent as 22-bit, two's complement signed integers.

This seems to fix the reported location of several stations, although there are still a few that still report bogus locations. However, as far as I can tell, this is a problem with those stations -- the raw bits do not make any sense when manually decoded.

### DIFF
--- a/src/pids.c
+++ b/src/pids.c
@@ -94,6 +94,7 @@ void decode_sis(pids_t *st, uint8_t *bits)
         int seq;
         int current_frame;
         int last_frame;
+        int fixed_pt_pos;
         float latitude, longitude;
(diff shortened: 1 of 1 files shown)


argilo:
Thanks @supersat! I checked the recordings I made in a bunch of U.S. cities, and many of them now show the correct location. (A bunch show the bogus location 99.214600, -11.968628, 128m, which I assume is a factory default.) So it would appear two's complement is the correct way to interpret the location.

PR #141, opened 2019-01-24
Fix memory leaks.

argilo:
After noticing a memory leak in #140 I decided to check for others. Using valgrind, I found three more. These changes should fix them.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -608,6 +608,7 @@ static void process_port(output_t *st, uint16_t port_id, uint8_t *buf, unsigned 
             // unk (4 bytes)
             // unk (4 bytes)
             port->u.file.size = p32[2];
+            free(port->u.file.data);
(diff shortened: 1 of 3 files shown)


src/output.c
```diff
@@ -608,6 +608,7 @@ static void process_port(output_t *st, uint16_t port_id, uint8_t *buf, unsigned
             // unk (4 bytes)
             // unk (4 bytes)
             port->u.file.size = p32[2];
+            free(port->u.file.data);
```
argilo: This matches the pattern used for `port->u.file.name` on lines 629 & 630.

awesie:
These look good. If stack usage becomes an issue, an easier solution is to modify main.c to allocate input_t on the heap rather than keep track of the additional heap allocations in frame_t.

PR #142, opened 2019-01-24
Move large stack allocations to the heap.

argilo:
@awesie's comment in https://github.com/theori-io/nrsc5/pull/141#issuecomment-457061952 prompted me to check for large stack allocations (with `-fstack-usage`). The worst offender by far was `uint8_t tmp[RADIO_BUFFER];` at 512 kB, followed by `input_t input;` at 52 kB. I've changed these both to heap allocations, and threw in `output_t output;` (2 kB) with them.

The program now happily runs with `ulimit -s 128`.


argilo:
It does care, but almost none of the one-time heap allocations in the program are freed, so I didn't bother. I do see valgrind complaining about use of uninitialized memory though, so I'll investigate that and update the PR if appropriate.

argilo:
Turns out the errors were there even before this PR, but may as well fix them now. :smile:

PR #143, opened 2019-01-26
Free allocated memory and resources.

argilo:
As noted in https://github.com/theori-io/nrsc5/pull/142, nrsc5 generally does not free its one-time memory allocations. That's fine, since the operating system frees them when the process ends, but it makes it much harder to detect true memory leaks with valgrind.

To solve the problem, I've rolled many of the memory allocations into arrays in the `input_t` and `output_t` structs, added `free`s where they were missing, and added cleanup code for other resources (files, FIR filters, libao, etc.)

Unfortunately, libao doesn't seem to free all its resources even when `ao_close` and `ao_shutdown` are called. But when non-ao output modes are used, valgrind reports zero heap allocations at exit:

```
$ valgrind --leak-check=full --show-leak-kinds=all src/nrsc5 -r cjmj2.raw -o /dev/null -f hdc

...

==13205== 
==13205== HEAP SUMMARY:
==13205==     in use at exit: 0 bytes in 0 blocks
==13205==   total heap usage: 46,697 allocs, 46,697 frees, 1,295,341,459 bytes allocated
==13205== 
==13205== All heap blocks were freed -- no leaks are possible
==13205== 
==13205== For counts of detected and suppressed errors, rerun with: -v
==13205== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

This should make it much easier to find memory leaks (such as those found in https://github.com/theori-io/nrsc5/pull/140 and https://github.com/theori-io/nrsc5/pull/141) in the future.


argilo:
I just noticed the program hangs at exit when WAV output is selected. Fixing that now...

argilo:
The deadlock should be fixed now.

PR #144, opened 2019-01-26
Use memcmp to check BBM.

argilo:
This change addresses the following compiler warning:

```
/home/travis/build/argilo/nrsc5/src/frame.c: In function ‘process_fixed_data’:
/home/travis/build/argilo/nrsc5/src/frame.c:434:13: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
             if (subch->block_idx == 4 && *(uint32_t *)subch->blocks != BBM)
             ^
```

I expect the old code would only work on little-endian systems.

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -24,7 +24,6 @@
 #define PCI_AUDIO_FIXED 0xE3634C
 #define PCI_AUDIO_FIXED_OPP 0x8D8D33
 
-#define BBM 0x42E23A7D
 #define MAX_AUDIO_PACKETS 64
 
 typedef struct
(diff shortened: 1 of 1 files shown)


src/frame.c
```diff
@@ -24,9 +24,10 @@
 #define PCI_AUDIO_FIXED 0xE3634C
 #define PCI_AUDIO_FIXED_OPP 0x8D8D33
 
-#define BBM 0x42E23A7D
 #define MAX_AUDIO_PACKETS 64
 
+static uint8_t bbm[] = { 0x7D, 0x3A, 0xE2, 0x42 };
```
awesie: Add a const. Also you may want to consider moving this into process_fixed_data as the first line since it is only used in that function.

argilo: Done.

PR #146, opened 2019-01-27
Address compiler warnings.

argilo:
I turned on `-Wall -Wextra -Wpedantic` to see what was lurking and fixed everything that came up: unused arguments and functions, signed/unsigned comparisons, and use of GNU extensions.


argilo:
The other gcc-ism was the multi-statement `MIN` macro defined in reed-solomon.c. The macro was only used once and its arguments didn't have side effects, so I replaced it with a simpler macro.

PR #131, opened 2018-02-19
Refactor into a library and use SoapySDR.

awesie:
Major change to provide a public API that is used by the
example command line application.

Fix #122 by using SoapySDR instead of librtlsdr.

PR #147, opened 2019-01-27
Refactor into a library

awesie:
Based on #131.

There are some minor changes to the arguments for the CLI app, but I tried to keep it mostly the same. Notably, the gain should be provided in dB instead of tenths of a dB. Also support for ADTS output is removed since it has been missing some HDC support for a while.

The public API is in include/nrsc5.h. Comments from anyone that would want to build an app around it are welcome.


argilo:
Thanks for taking another crack at this! I ran a quick test to verify that the CLI app can still receive my local stations. I'll have a closer look at the API and code changes now.

argilo:
A few random thoughts based on what I've seen so far:

* Since HD Radio lives between 87.9 and 107.9 MHz and station frequencies are advertised in MHz, maybe it would make sense to have MHz be the default input format for the CLI app, and switch over to Hz if the value is greater than a threshold.
* Since programs 0-7 are presented to the user as HD1 through HD8 on radios, would it make more sense to use 1-8 as the program argument in the CLI app?
* Should the API include an option to read IQ samples through a buffer?
* If so, the RTL-SDR & file-reading code could even move outside the library, although that would leave more work to the app developer.

I'm not sure if these are good ideas or not, just bringing them up for discussion.

argilo:
Another argument for keeping the RTL-SDR code inside the library is that it allows the library to handle AGC.

src/main.c
```diff
@@ -13,134 +13,332 @@
...
-    output_t *output;
-
-    while ((opt = getopt_long(argc, argv, "r:w:d:p:o:f:g:ql:v", long_opts, NULL)) != -1)
+    char *output_name = NULL, *audio_name = NULL, *hdc_name = NULL;
+    char *endptr;
+    int opt;
+
+    st->gain = -1;
+
+    while ((opt = getopt_long(argc, argv, "r:w:o:d:g:ql:v", long_opts, NULL)) != -1)
```
argilo:
```suggestion
    while ((opt = getopt_long(argc, argv, "r:w:o:d:p:g:ql:v", long_opts, NULL)) != -1)
```
This fixes the `-p` option.

awesie: Oops. Thanks! Hopefully there aren't too many similar things from try to pry these PRs apart.

awesie:
Re your comments:

- The CLI as-is should support this behavior. Any frequency less than 10,000 will be interpreted in MHz.
- I could be convinced. A related thing would be to make the parameter optional and default to 0 (HD1).
- A reasonable idea. I could remove the file reading code from the nrsc5 library and move that entirely into the CLI. Would we continue to use rtlsdr format for the IQ file?
- Oops, got ahead of myself. I agree with moving the file related code, but I think the library should continue to handle interactions with the radio hardware. The API should make it easy to make an app that "just works", and if we provide an API for advanced use cases that want to give us raw IQ, great, but that should not be the default.

Another thing that might make sense is to reorganize the repository, if we would ever want to have multiple apps in tree. Though maybe the right answer is for individual apps to live in separate repositories.

I also haven't looked at all at what it would take to integrate this library with any existing SDR apps. If anyone has any thoughts on this, e.g. has the desire and expertise to do it, I'd be interesting in hearing about it. My opinion is that this is unnecessary and it would be better to just build a simple (HD) FM radio GUI, but I am not really an end user.

awesie:
For one possible way of handling multiple apps, I ported my web server commit to this branch (https://github.com/awesie/nrsc5/compare/library-refactor...awesie:webserver-app). It moves the CLI to app/cli/ and web server to app/web/.

There are also a couple of other patches in that branch to provide some quick SIS information to the public API, move logging output to the app, and reset the various decoding structures whenever the change frequency API is called. I'll probably add these other commits to this PR but I wanted to wait on feedback on the current PR as-is.

argilo:
> The CLI as-is should support this behavior. Any frequency less than 10,000 will be interpreted in MHz.

That's true. I noticed the readme examples switched to Hz though, which suggests it's now the preferred format.

> A related thing would be to make the parameter optional and default to 0 (HD1).

I like that idea. Receivers default to HD1, and I've forgotten the `0` parameter on many occasions.

argilo:
> I could remove the file reading code from the nrsc5 library and move that entirely into the CLI. Would we continue to use rtlsdr format for the IQ file?

I like the idea of moving the file-reading code out into the CLI. i don't have a strong preference on file format. The RTL-SDR format (8-bit unsigned) is space-efficient and compatible with the `rtl_sdr` command line tool, but used only for RTL-SDR. 32-bit float is commonly used in the GNU Radio world, and would probably be a sensible choice for the API, to take advantage of the higher resolution of other SDRs.

it might make sense to leave such a change for a future PR.

> I also haven't looked at all at what it would take to integrate this library with any existing SDR apps.

The one I had in mind was GNU Radio. If the nrsc5 API supported direct IQ input, then it would be simple to wrap it in a GNU Radio block. it could then be integrated into gqrx or used in custom flow graphs.

argilo:
A minor bug: If the `-r` and `-o` options are used together, then a deadlock occurs after the input file is read. The deadlock occurs at these two lines:

https://github.com/awesie/nrsc5/blob/ce7387d187b42c70e2d9f9aad9d73e392dfebd51/src/main.c#L481
https://github.com/awesie/nrsc5/blob/ce7387d187b42c70e2d9f9aad9d73e392dfebd51/src/nrsc5.c#L108

include/nrsc5.h
```diff
@@ -0,0 +1,173 @@
...
+int nrsc5_open_iq(nrsc5_t **, const char *path);
+void nrsc5_close(nrsc5_t *);
+void nrsc5_start(nrsc5_t *);
+void nrsc5_stop(nrsc5_t *);
+void nrsc5_get_frequency(nrsc5_t *, float *freq);
+int nrsc5_set_frequency(nrsc5_t *, float freq);
+void nrsc5_get_gain(nrsc5_t *, float *gain);
+int nrsc5_set_gain(nrsc5_t *, float gain);
+void nrsc5_set_auto_gain(nrsc5_t *, int enabled);
+int nrsc5_scan(nrsc5_t *, float begin, float end, float skip, float *result);
```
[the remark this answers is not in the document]
awesie: It was on my list of features that I would want the API to expose. I'll remove it from this PR and wait until an implementation (plus the prototype is missing a name result argument anyways).

awesie:
> > The CLI as-is should support this behavior. Any frequency less than 10,000 will be interpreted in MHz.
> 
> That's true. I noticed the readme examples switched to Hz though, which suggests it's now the preferred format.
> 

You are probably right that I should leave the README alone. For some reason seeing a station frequency without anything after (M, e6) was giving me an itch.

> 
> The one I had in mind was GNU Radio. If the nrsc5 API supported direct IQ input, then it would be simple to wrap it in a GNU Radio block. it could then be integrated into gqrx or used in custom flow graphs.

For the direct IQ input, I assume that it would be easiest for the external app to push samples to the nrsc5 library, as opposed to pull via a callback? At the moment, reading samples from the input and processing the samples is all on one worker thread. Actually, easiest thing would be to use pipe() internally.

argilo:
> I assume that it would be easiest for the external app to push samples to the nrsc5 library, as opposed to pull via a callback?

Yes, that's the best option for the use case I had in mind. The GNU Radio scheduler pushes samples into signal processing blocks, so it would be convenient for an nrsc5 block to then push the samples into libnrsc5.

src/nrsc5.c
```diff
@@ -0,0 +1,516 @@
...
+        while (!st->auto_gain_snr_ready)
+        {
+            int len = sizeof(st->samples_buf);
+
+            if (rtlsdr_read_sync(st->dev, st->samples_buf, len, &len) != 0)
+                goto error;
+
+            input_cb(st->samples_buf, len, &st->input);
+        }
+        log_debug("AGC: %.2f (%.2f)", gain / 10.0f, st->auto_gain_snr);
```
argilo: Nitpick: I'd prefer to keep the "Gain" and "CNR" labels in place here (and below on line 61). The old format was `"Gain: %.1f dB, CNR: %.1f dB"`.

src/private.h
```diff
@@ -0,0 +1,49 @@
...
+#include "input.h"
+#include "output.h"
+
+struct nrsc5_t
+{
+    rtlsdr_dev_t *dev;
+    FILE *iq_file;
+    uint8_t samples_buf[128 * 256];
+    float freq;
+    int decimation;
```
argilo: This seems unused. `nrsc5_open_iq` sets it to 2, and that's it.

include/nrsc5.h
```diff
@@ -0,0 +1,173 @@
...
+    NRSC5_EVENT_SYNC,
+    NRSC5_EVENT_LOST_SYNC,
+    NRSC5_EVENT_MER,
+    NRSC5_EVENT_BER,
+    NRSC5_EVENT_HDC,
+    NRSC5_EVENT_AUDIO,
+    NRSC5_EVENT_SIS,
+    NRSC5_EVENT_ID3,
+    NRSC5_EVENT_SIG,
+    NRSC5_EVENT_LOT,
```
argilo: Out of curiosity, what's the meaning of LOT?

include/nrsc5.h
```diff
@@ -0,0 +1,173 @@
...
+    NRSC5_EVENT_TRAFFIC,
+    NRSC5_EVENT_WEATHER
```
argilo: The traffic and weather events are unused. Did you have plans for them?

src/output.h
```diff
@@ -2,91 +2,137 @@
...
-    OUTPUT_WAV,
-    OUTPUT_LIVE
-} output_method_t;
+    NonSpecific = 0,
+    News = 1,
+    Sports = 3,
+    Weather = 29,
+    Emergency = 31,
+    Traffic = 65,
+    ImageMaps = 66,
```
argilo: The table in 1020s also includes Text = 80.

src/output.c
```diff
@@ -540,108 +273,220 @@ static void output_id3(uint8_t *buf, unsigned int len)
...
+{
+    switch (type)
+    {
+    case NonSpecific: return "Non-specific";
+    case News: return "News";
+    case Sports: return "Sports";
+    case Weather: return "Weather";
+    case Emergency: return "Emergency";
+    case Traffic: return "Traffic";
+    case ImageMaps: return "Image Maps";
```
argilo: If Text is added to output.h it will need to go here too.

src/output.c
```diff
@@ -650,78 +495,173 @@ static void process_port(output_t *st, uint16_t port_id, uint8_t *buf, unsigned
...
     {
-        uint32_t seq = buf[4] | (buf[5] << 8) | (buf[6] << 16) | (buf[7] << 24);
+        if (len < 8)
+        {
+            log_warn("bad fragment (port %04X, len %d)", port_id, len);
+            return;
+        }
+        uint8_t hdrlen = buf[0];
+        uint16_t lot = *(uint16_t*)&buf[2];
+        uint32_t seq = *(uint32_t*)&buf[4];
```
argilo: This went back to assuming little-endian.

src/output.c
```diff
@@ -650,78 +495,173 @@ static void process_port(output_t *st, uint16_t port_id, uint8_t *buf, unsigned
...
-            free(port->u.file.name);
-            port->u.file.name = strndup((const char *)buf, namelen);
-            buf += namelen;
-            len -= namelen;
+            // uint32_t == 1
+            // uint32_t xxx
+            // uint32_t size
+            // uint32_t mimeHash
+            file->size = *(uint32_t*)&buf[8];
+            file->mime = *(uint32_t*)&buf[12];
```
argilo: More little-endian here.

argilo:
I've finished a first pass over the code changes. There's a few minor issues to be fixed here and there, but the overall approach looks good and I like the design of the API. I wouldn't be against merging this soon and carrying on with the other changes you mentioned (logging, SIS, etc.) in subsequent PRs.

awesie:
Thanks! I'll work through the minor fixes and address some of the comments individually.

pclov3r:
@awesie Will you Webserver eventually come to the main branch or what is your ideas for that? Sounds like it would be rather useful.

pclov3r:
Also would like to add so far this PR doesn't seem to have changed anything for Wndows. Still functioning fine.

awesie:
@pclov3r Thanks for testing.

Re the web server app, my hope is that it will be integrated at some point. It won't be until after this PR and a couple of follow up PRs are merged.

include/nrsc5.h
awesie: When I designed the API, it was before I started implementing things. These were non-trivial and not useful enough to warrant the effort. I'll remove the define until we have an implementation.

src/private.h
awesie: Yes, left over from the SoapySDR support.

PR #151, opened 2019-01-30
Add support for grabbing samples over a pipe.

awesie:
This adds two new functions: nrsc5_open_pipe and nrsc5_pipe_samples. It also
replaces nrsc5_open_iq with nrsc5_open_fd. As part of this change, we no longer
use stdio functions but open, read and write.


argilo:
The `-r` option in the CLI still works for me. I haven't tried the pipe option yet, but will do that tomorrow morning.

PR #156, opened 2019-02-02
Use _pipe instead of pipe on Windows.

argilo:
MinGW doesn't have a `pipe` function, so nrsc5 fails to build on Windows. Using `_pipe` solves the problem. If there's a better way to do this, please let me know.

I tested the API in Python to verify that pipe input works.

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -5,6 +5,11 @@
 
 #include "private.h"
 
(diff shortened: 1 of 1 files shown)


argilo:
I chose 65536 as the size since that appears to be the default on Linux.

PR #157, opened 2019-02-02
Make libnrsc5 a shared library.

argilo:
@awesie What do you think about building libnrsc5 as a shared library? This makes it possible to load it in Python using `ctypes`.


argilo:
OK, I'll take a crack at building both and update this PR.

argilo:
@awesie I've updated this to produce both static and shared libraries, and to have the CLI link to the static library.

PR #158, opened 2019-02-02
Add a build script for MSYS2.

argilo:
Following the build instructions for Windows is tedious. To simplify the process, I moved as much as I could into a bash script. The script also makes a couple improvements:

* The MinGW environment variables are used to determine which compiler & prefix should be used. This allows both 32-bit and 64-bit builds.
* FFTW is updated to version 3.3.8.
* For dependencies fetched from git, the latest release tag is used (instead of the master branch), which hopefully will reduce the likelihood of bugs.

The script tries to avoid repeating work by checking whether the dependencies are already downloaded or installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,53 +88,16 @@ The only build environment that has been tested on Windows is MSYS2 with MinGW. 
 
 ### Building with [MSYS2](http://www.msys2.org)
 
-Install MSYS2. Open a terminal using the "MSYS2 MinGW 32-bit" shortcut.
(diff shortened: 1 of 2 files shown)


argilo:
@pclov3r You may be interested in testing this out.

pclov3r:
Somebody else will have to test this. I cross compile it for windows from Debian and don't have a mysis2 setup here.

argilo:
No problem.

As it happens, I was just experimenting with cross-compiling myself. I'd be happy to hear your feedback on the approach I took in https://github.com/theori-io/nrsc5/pull/159.

PR #159, opened 2019-02-03
Add script to cross-compile for Windows.

argilo:
I've put together a shell script (`support/win-cross-compile`) to cross-compile Windows binaries on Linux. This has the advantage of being much faster than MSYS2, and allows Windows builds to be tested in Travis CI, which should help catch bugs like https://github.com/theori-io/nrsc5/pull/156 earlier.

To run the script, first install the MinGW compiler with `sudo apt install mingw-w64`, then run `support/win-cross-compile 32` to build 32-bit binaries or `support/win-cross-compile 64` to build 64-bit binaries. The output files will be in `build-win32/bin` or `build-win64/bin`.

I've tested this on Ubuntu 16.04 (through Travis CI) and Ubuntu 18.04. @pclov3r I'd be interested to hear if this works on Debian for you.

I'll add some documentation to the readme soon.

### DIFF
--- a/support/win-cross-compile
+++ b/support/win-cross-compile
@@ -76,4 +76,9 @@ C_INCLUDE_PATH=${prefix}/include make
 make install
 
 cp /usr/${host}/lib/libwinpthread-1.dll ${prefix}/bin
-cp /usr/lib/gcc/${host}/*-win32/libgcc_s_*-1.dll ${prefix}/bin
+
+if [ "$1" == 32 ]; then
+    cp "$(${host}-gcc -print-file-name=libgcc_s_sjlj-1.dll)" "${prefix}/bin"
+elif [ "$1" == 64 ]; then
+    cp "$(${host}-gcc -print-file-name=libgcc_s_seh-1.dll)" "${prefix}/bin"
+fi
--- a/support/win-cross-compile
+++ b/support/win-cross-compile
@@ -1,0 +1,79 @@
+#!/usr/bin/env bash
+
+set -e
+
+fftw_version=3.3.8
+libao_version=1.2.2
+libusb_version=v1.0.22
+rtlsdr_version=0.6.0
+
+root=`git rev-parse --show-toplevel`
+if [ "$1" == 32 ]; then
+    prefix=${root}/build-win32
+    host=i686-w64-mingw32
+elif [ "$1" == 64 ]; then
+    prefix=${root}/build-win64
+    host=x86_64-w64-mingw32
+else
+    echo "Usage: $0 (32|64)"
+    exit 1
+fi
+
+mkdir -p ${prefix}
+
+cd ${prefix}
+if [ ! -d fftw-${fftw_version} ]; then
+    curl -L http://www.fftw.org/fftw-${fftw_version}.tar.gz | tar xz
+fi
+if [ ! -e ${prefix}/lib/libfftw3f.a ]; then
+    cd fftw-${fftw_version}
+    ./configure --host=${host} --prefix=${prefix} --enable-float --enable-sse2 --with-our-malloc
+    make
+    make install
+fi
+
+cd ${prefix}
+if [ ! -d libao ]; then
+    git clone https://git.xiph.org/libao.git
+fi
+if [ ! -e ${prefix}/bin/libao-4.dll ]; then
+    cd libao
+    git checkout ${libao_version}
+    ./autogen.sh
+    LDFLAGS=-lksuser ./configure --host=${host} --prefix=${prefix} --enable-wmm --disable-pulse
+    make
+    make install
+fi
+
+cd ${prefix}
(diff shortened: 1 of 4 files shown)


pclov3r:
One suggestion i just had is adding a option to compile libfftw with AVX support. Alot of new modern CPUs over the past several years support this.

pclov3r:
Just reading the documentation it appears it's safe to include this 

> Enable various SIMD instruction sets. You need compiler that supports the given SIMD extensions, but FFTW will try to detect at runtime whether the CPU supports these extensions. That is, you can compile with--enable-avx and the code will still run on a CPU without AVX support http://fftw.org/fftw3_doc/Installation-on-Unix.html

Guess it applies to Windows builds as well.

awesie:
I'll test these commits today, but it looks good. Future work might be to build the dependencies using cmake if system libraries do not exist, similar to what some other projects do. This script is more than sufficient for now.

argilo:
> Future work might be to build the dependencies using cmake if system libraries do not exist, similar to what some other projects do.

Agreed, that would probably be a cleaner way to do it.

For now I just wanted something quick & dirty that would save me from waiting 40 minutes for MSYS2 to compile, and allow Windows builds to be tested in Travis CI. (Travis CI actually supports Windows builds, but the compiler is Visual Studio.)

support/win-cross-compile
```diff
@@ -0,0 +1,79 @@
...
+    make
+    make install
+fi
+
+cd ${prefix}
+cmake -D CMAKE_SYSTEM_NAME=Windows -D CMAKE_C_COMPILER=${host}-gcc -D CMAKE_LIBRARY_PATH=${prefix}/lib -D FAAD2_CONFIGURE_ARGS="--host=${host}" -D USE_COLOR=OFF -D USE_SSE=ON -D CMAKE_INSTALL_PREFIX=${prefix} ..
+C_INCLUDE_PATH=${prefix}/include make
+make install
+
+cp /usr/${host}/lib/libwinpthread-1.dll ${prefix}/bin
```
awesie:
```suggestion
# On MacOS, libwinpthread-1.dll is in /bin so we need to use -print-prog-name.
# On Linux, libwinpthread-1.dll is in /lib so we need to use -print-file-name.
LIBWINPTHREAD_PATH="$(${host}-gcc -print-prog-name=libwinpthread-1.dll)"
if [ ! -e "${LIBWINPTHREAD_PATH}" ]; then
    LIBWINPTHREAD_PATH="$(${host}-gcc -print-file-name=libwinpthread-1.dll)"
fi
cp "${LIBWINPTHREAD_PATH}" "${prefix}/bin"
```

support/win-cross-compile
```diff
@@ -0,0 +1,79 @@
...
+cp /usr/lib/gcc/${host}/*-win32/libgcc_s_*-1.dll ${prefix}/bin
```
awesie:
```suggestion

if [ "$1" == 32 ]; then
    cp "$(${host}-gcc -print-file-name=libgcc_s_sjlj-1.dll)" "${prefix}/bin"
elif [ "$1" == 64 ]; then
    cp "$(${host}-gcc -print-file-name=libgcc_s_seh-1.dll)" "${prefix}/bin"
fi
```

argilo:
What's the brew command to install the compiler? I could even try adding that into CI to make sure it stays functional.

awesie:
It looks like it should be ```brew install mingw-w64```

argilo:
Thanks. For now I'll merge this, then play with CI and open another PR if it works.

awesie:
```
  Backends to be built:

    AU file output: .............. yes
    RAW file output: ............. yes
    WAV file output: ............. yes

    AIX live output: ............. no
    ALSA live output: ............ no
    ARTS live output: ............ no
    ESD live output: ............. no
    IRIX live output: ............ no
    MACOSX live output: .......... no
    NAS live output: ............. no
    NULL live output: ............ yes
    OSS live output: ............. no
    PULSE live output: ........... no
    ROAR live output: ............ no
    SNDIO live output: ........... no
    SUN live output: ............. no
    Windows WMM live output: ..... yes
```

argilo:
Perfect, thanks!

pclov3r:
I just thought of something to add to this. Is it possible to add a option to define the number of threads to use? So for example you could do something like ./win-cross-compile 64 8 for 64-bit arch and 8 threads? 

Unsure if defining the number of threads makes a difference now for compile speed.

pclov3r:
> I just thought of something to add to this. Is it possible to add a option to define the number of threads to use? So for example you could do something like ./win-cross-compile 64 8 for 64-bit arch and 8 threads?
> 
> Unsure if defining the number of threads makes a difference now for compile speed.

It seems https://github.com/theori-io/nrsc5/pull/170 added this. Thanks!

PR #162, opened 2019-02-03
Switch back to stdio for file reading.

### LINKED ISSUE #160: IQ input is broken on Windows
When the `-r` option is used on Windows, nrsc5 stops reading IQ input as soon as the first 0x1A byte is reached. This happens because it's interpreted as the end-of-file marker.

Adding the `O_BINARY` flag to the `open` call here solves the problem (for file input, but not stdin):

https://github.com/theori-io/nrsc5/blob/c351235f12a61f961632d1090416b28434407058/src/main.c#L477

But that flag is only available in MinGW, so we'd need another `#ifdef`. Maybe it's best of avoid using POSIX calls.

### LINKED ISSUE #161: Assertion failure in input_push
The `input_push` function expects its `len` argument to be a multiple of four:

https://github.com/theori-io/nrsc5/blob/c351235f12a61f961632d1090416b28434407058/src/input.c#L154

But `worker_thread` calls it without making sure that's the case:

https://github.com/theori-io/nrsc5/blob/c351235f12a61f961632d1090416b28434407058/src/nrsc5.c#L138-L140

This can happen if the length of the input file is not a multiple of four bytes:
```
$ echo -n "a" | src/nrsc5 -r - 0
nrsc5: /home/argilo/git/nrsc5/src/input.c:154: input_push: Assertion `len % 4 == 0' failed.
Aborted (core dumped)
```
But I suppose it could also happen if the `read` function was interrupted by a system call.

Previously I solved this problem by reading bytes four at a time: https://github.com/theori-io/nrsc5/pull/62

But that's not possible with `read` so we'd have to handle it ourselves. Maybe this is another good argument for getting rid of POSIX calls and going back to the standard C library.

argilo:
Fixes #160. Fixes #161.

This change switches from POSIX calls back to the standard C library for file input. I changed the API to accept a file pointer instead of a file descriptor. Things appear to be working well on Windows now, and I tested to make sure "pipe" input is still working.

Getting rid of the pipe also saves us an ifdef (from #156) and eliminates system calls.

I guess it's worth considering whether `nrsc5_open_pipe` and `nrsc5_pipe_samples` are still sensible names, since the underlying pipe was just an implementation detail. I guess they do still let you pipe in samples, in some sense.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -6,6 +6,7 @@
  */
 #include <stdint.h>
 #include <stdlib.h>
+#include <stdio.h>
 
 /*
  * Definitions.
@@ -155,7 +156,7 @@ typedef struct nrsc5_t nrsc5_t;
  * Public functions. All functions return void or an error code (0 == success).
  */
 int nrsc5_open(nrsc5_t **, int device_index, int ppm_error);
-int nrsc5_open_fd(nrsc5_t **, int fd);
+int nrsc5_open_file(nrsc5_t **, FILE *fp);
 int nrsc5_open_pipe(nrsc5_t **);
 void nrsc5_close(nrsc5_t *);
 void nrsc5_start(nrsc5_t *);
--- a/src/main.c
+++ b/src/main.c
@@ -14,12 +14,10 @@
  */
 
 #include <ao/ao.h>
-#include <fcntl.h>
 #include <getopt.h>
 #include <nrsc5.h>
 #include <pthread.h>
 #include <sys/time.h>
-#include <unistd.h>
 
 #include <assert.h>
 #include <stdio.h>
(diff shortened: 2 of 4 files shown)


awesie:
I just didn't have a better name at the time. Any thoughts on that?

Also the reason I didn't initially just call input_push from pipe_samples is because I was trying to keep the processing of the samples on a separate thread from what is reading the samples. For instance, input_push may block for a bit causing the caller of pipe_samples to drop samples. Thoughts?

argilo:
For naming, "stream" in place of "pipe" crossed my mind, but I think I like the current name better.

> `input_push` may block for a bit causing the caller of pipe_samples to drop samples

I guess it's true that the pipe did include a buffer (64 kB by default on Linux), but you'd still go back to blocking once that was full.

I don't have a strong opinion here. For the one use case I had in mind (a GNU Radio block), blocking won't be an issue.

awesie:
If keeping pipe is useful, we could wrap the file descriptors in a buffered stdio object using fdopen (posix) and _fdopen (windows). But you might be right that the pipe buffer was too small for it to be useful.

argilo:
> If we need to [...] add buffering in the future, the current API gives us the flexibility to do that.

Yep, and of course that could be implemented by something other than a pipe if need be.

PR #163, opened 2019-02-03
Prototype a Python CLI app.

argilo:
It might be useful to merge in my prototype Python CLI app at this point. it's already helping to test the API, and it can be executed as part of Travis CI so help us catch bugs earlier. (For instance in `nrsc5_open_pipe` and `nrsc5_pipe_samples`, which are not used by the C CLI app.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,9 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
 
 install:
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -y libao-dev libfftw3-dev librtlsdr-dev mingw-w64; fi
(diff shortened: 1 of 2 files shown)


argilo:
Heres the result of the CI run: https://travis-ci.org/argilo/nrsc5/builds/488195677

pclov3r:
Is there something i'm missing here? When attempting to use this i get ImportError: No module named 'pyaudio' however it's installed.

Any pointers?

pclov3r:
My fault scratch that. Needed to install the python3 version of pyaudio.

argilo:
In case it's useful to you, I got stuck trying to compile it and used the binary available from here: https://www.lfd.uci.edu/~gohlke/pythonlibs/#pyaudio

argilo:
By the way, here's what the Python CLI can do so far:

* take input from RTL-SDR, file, or stdin
* play audio packets on the sound card

And that's about it! It currently ignores all the other events produced by libnrsc5.

It might be nice to split it into two pieces: a Python wrapper for libnrsc5, and the CLI. Other apps could reuse the Python wrapper and make their own UI.

PR #166, opened 2019-02-04
Support character encodings in SIS and PSD.

argilo:
Both SIS messages and ID3 tags in PSD have a character encoding associated with them, either ISO-8859-1 or UCS-2. So far nrsc5 has been ignoring the coding and printing out the raw bytes. This PR adds functions to convert from these encodings to UTF-8, and calls them where appropriate.

I was able to test ISO-8859-1 in ID3 directly because a local station sometimes uses non-ASCII characters. For the other cases I was only able to verify that ASCII messages were still displayed correctly. I haven't observed any stations using UCS-2.

I tested the `iso_8859_1_to_utf_8` and `ucs_2_to_utf_8` in a stand-alone program to verify that they work properly.

The `ucs_2_to_utf_8` function assumes little-endian when there is no Byte Order Mark since SIS specifies little-endian for UCS-2. ID3 can use big-endian or little-endian for UCS-2, but the Byte Order Mark is mandatory there.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,6 +29,7 @@ add_library (
     galois.c
 
     log.c
+    unicode.c
 
     strndup.c
 )
--- a/src/main.c
+++ b/src/main.c
@@ -23,6 +23,10 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifdef __MINGW32__
+#include <windows.h>
+#endif
+
 #include "bitwriter.h"
 #include "log.h"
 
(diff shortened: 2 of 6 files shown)


argilo:
The only problem I noticed so far is that Unicode still doesn't work on Windows. Apparently you need to do some special things to declare Unicode support in MinGW. I'll open another PR for that if I get it working.

argilo:
`SetConsoleOutputCP(CP_UTF8)` seems to do the trick. I've added it into this PR.

PR #167, opened 2019-02-04
Enable AVX when building FFTW on Windows.

argilo:
As @pclov3r suggested in https://github.com/theori-io/nrsc5/pull/159, I've added the `--enable-avx` option to FFTW when building it for Windows. I did some timings with PowerShell's `Measure-Command` and it seems to give a speedup of around 3% on my laptop. Debian also compiles its FFTW package with `--enable-avx` so that's another good indication it's safe to use.

### DIFF
--- a/support/msys2-build
+++ b/support/msys2-build
@@ -16,7 +16,7 @@ if [ ! -d fftw-${fftw_version} ]; then
 fi
 if [ ! -e ${MINGW_PREFIX}/lib/libfftw3f.a ]; then
     cd fftw-${fftw_version}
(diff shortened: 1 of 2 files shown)


argilo:
The laptop I tested with does not have AVX2, so i didn't try that. Do you have a machine to test with? i used this PowerShell command to measure:

`Measure-Command { .\nrsc5.exe -r samples.dat -o NUL 0 }`

pclov3r:
@argilo I do i'll give it a go later. AMD R7 2700x here tho.

pclov3r:
It appears to maybe be slightly better but the difference isn't huge. At best maybe 2%.

PR #168, opened 2019-02-06
Avoid stepping into fixed data when processing audio payload.

argilo:
I noticed occasional "RS corrected 5 symbols" log messages even when listening to stations with a BER of zero, so I dug in to see what was going wrong. The problem is that the main loop in `frame_process` continues until it reaches the end of the PDU. If the PDU ends with fixed data, then the loop attempts to process it as if it was an audio packet. Usually this causes Reed-Solomon error correction to fail, but occasionally it succeeds and prints the "RS corrected 5 symbols" message. Audio packet processing then proceeds with bogus data.

To solve the problem I modified `process_fixed_data` to return the length of the PDU minus the amount of bytes it processed. Once it determines the lengths of the fixed data streams, the excess processing of audio packets stops.

Along the way I renamed `PDU_LEN` to `MAX_PDU_LEN` to better reflect its meaning, and added the PDU length as an argument to `process_fixed_data` rather than having it assume it's processing a PDU from the P1 channel.

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -38,7 +38,7 @@
 // bits per L2 PCI
 #define PCI_LEN 24
 // bytes per L2 PDU (max)
-#define PDU_LEN ((P1_FRAME_LEN - PCI_LEN) / 8)
+#define MAX_PDU_LEN ((P1_FRAME_LEN - PCI_LEN) / 8)
 // number of programs (max)
 #define MAX_PROGRAMS 8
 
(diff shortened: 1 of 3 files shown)


awesie:
The patch looks okay.

I am surprised that the extra data gets past both the RS and CRC-8. I wonder if it would make sense to make fix_header a little more robust, e.g. verify that the padding zero bytes remain zero after rs_decode, don't allow more than E (4) corrections, etc.

argilo:
I'm actually looking into the Reed-Solomon error correction now. It seems to be broken, because if I intentionally introduce one error it can't correct it. And it only ever reports correcting five errors, which should be impossible. I'll open up another PR once I sort out what's wrong with it. I was also planning to look into how to make this more robust, as the length of the fixed data is not known until a few PDUs arrive.

PR #170, opened 2019-02-06
Build dependencies using CMake

awesie:
This PR moves most of the logic to build external libraries to CMake instead of scripts. Its primarily targeted at cross compiling and building for Windows, since libao's dynamic plugins make it difficult to statically link on other platforms. I have not tested building with the built-in libraries on Linux yet.

There are also some changes to speed up the Travis CI build using ccache and parallel make jobs. It also tests the built Windows binary using wine on macOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,11 +4,24 @@ include (CheckSymbolExists)
 include (ExternalProject)
 project (nrsc5 C)
 
+execute_process (COMMAND ${CMAKE_C_COMPILER} -dumpmachine OUTPUT_VARIABLE HOST_TRIPLE_DEFAULT OUTPUT_STRIP_TRAILING_WHITESPACE)
+
 option (USE_COLOR "Colorize log output")
 option (USE_NEON "Use NEON instructions")
 option (USE_SSE "Use SSE3 instructions")
 option (USE_FAAD2 "AAC decoding with FAAD2" ON)
+option (USE_STATIC "Link with static libraries")
+option (USE_SYSTEM_FFTW "Use system provided fftw" ON)
+option (USE_SYSTEM_RTLSDR "Use system provided rtl-sdr" ON)
+option (USE_SYSTEM_LIBUSB "Use system provided libusb" ON)
+option (USE_SYSTEM_LIBAO "Use system provided libao" ON)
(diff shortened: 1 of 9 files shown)


argilo:
Nice! I had a quick look just now and there's a lot to like here. I'll do a more complete review tonight.

CMakeLists.txt
```diff
@@ -53,10 +92,119 @@ check_symbol_exists (CMPLXF complex.h HAVE_CMPLXF)
...
+
+    set (LIBAO_PREFIX "${CMAKE_BINARY_DIR}/libao-prefix")
+    ExternalProject_Add (
+        libao_external
+        GIT_REPOSITORY "https://git.xiph.org/libao.git"
+        GIT_TAG d5221655dfd1a2156aa6be83b5aadea7c1e0f5bd
+        PREFIX ${LIBAO_PREFIX}
+
+        UPDATE_COMMAND ""
+        CONFIGURE_COMMAND ${LIBAO_PREFIX}/src/libao_external/configure ${HOST_TRIPLE_ARG} --prefix=${LIBAO_PREFIX} --enable-static --disable-shared
```
argilo: It looks like `--disable-pulse` is needed here when cross-compiling on Linux.

argilo:
libnrsc5.dll depends on libwinpthread and libgcc, but they are not statically linked nor are they copied to the bin folder.

argilo:
The Windows instructions in the readme could use an update since only nrsc5.exe is needed once the build is finished. (And possibly libnrsc5.dll, for use with Python or other apps that might appear.)

argilo:
A couple minor things to be fixed up, but I really like these changes. Testing the Windows build in wine had crossed my mind, so I'm glad you added that to the Travis CI build. And statically linking everything into a single .exe file will be convenient for Windows users.

awesie:
This should resolve your comments. I felt more comfortable with statically linking the library once I made the exports explicit, so that the linker should hide the symbols from the static libraries.

Travis CI is building: https://travis-ci.org/awesie/nrsc5/builds/490333685.

argilo:
Nice! I'll take it for a spin and let you know how it goes.

argilo:
Works great now on Linux, and well as cross-compiling for Windows.

PR #171, opened 2019-02-07
Fix Reed-Solomon code and improve robustness.

argilo:
The Reed-Solomon error correction in frame.c was not working; it was unable to correct even a single corrupted symbol. I fought with it for a while, then gave up and replaced it with the Reed-Solomon code from GNU Radio (https://github.com/gnuradio/gnuradio/tree/master/gr-fec/lib/reed-solomon), originally written by Phil Karn. It's licensed under the GPL, so should be fine for this project.

I also made the robustness improvement suggested by @awesie in https://github.com/theori-io/nrsc5/pull/168#issuecomment-460927348; if Reed-Solomon corrects "errors" outside the codeword, then it is rejected. Prior to this improvement, about 4% of random inputs were accepted, and afterwards that went down to 0.08%.

I tested this change on my library of IQ files and observed some successful corrections of 1-4 symbols, as expected. The code was also able to correct 1-4 symbols when I purposely corrupted them.


awesie:
Looks good. gnu-radio is GPLv3 or later so license should not be a problem

PR #173, opened 2019-02-08
Start over from coarse sync if frame can't be decoded.

### LINKED ISSUE #172: Time synchronization in sync.c cannot tolerate dropped samples
I noticed that my Windows machine occasionally (once every few hours) stops receiving. The MER drops suddenly and never recovers. After some investigation, I discovered that the problem occurs when the RTL-SDR drops samples. The fine time synchronization code in `sync_process` can only tolerate a timing error of +/- 25 samples. Beyond that limit there is phase ambiguity in the reference subcarriers, and the synchronizer will happily lock onto the wrong phase.

The problem can easily be reproduced by removing bytes from the middle of an IQ file:

```
xz -d < ../support/sample.xz > ../support/sample
cat <(head -c 20000000 ../support/sample) <(tail -c +20000151 ../support/sample) | src/nrsc5 -r - 0
```

I'll have to think about a good way to detect this situation. The coarse time synchronization in `acquire_process` does not have this problem, so maybe it could be used to check if time synchronization has been lost.

argilo:
Fixes #172.

Since three components are now concerned with synchronization state, I've moved it out into the input struct. There are now three states:

* `SYNC_STATE_NONE` - not synchronized
* `SYNC_STATE_COARSE` - coarse synchronization has completed
* `SYNC_STATE_FINE` - fine synchronization has completed

If the reference subcarriers are not as expected or a frame fails to decode, then the state is set back to `SYNC_STATE_NONE`.

I'm still testing this out and would appreciate feedback. Testing instructions are available in #172. The amount of dropped samples can be changed by increasing or decreasing the number in the `tail` command. (Make sure the number remains odd, because each IQ sample is two bytes long.)

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -472,7 +472,11 @@ void frame_process(frame_t *st, size_t length)
         hef_t hef = {0};
 
         if (!fix_header(st, st->buffer + offset))
+        {
+            if (offset == 0)
+                st->input->sync_state = SYNC_STATE_NONE;
             return;
+        }
 
         parse_header(st->buffer + offset, &hdr);
         offset += 14;
--- a/src/frame.c
+++ b/src/frame.c
@@ -459,11 +459,12 @@ static size_t process_fixed_data(frame_t *st, size_t length)
 void frame_process(frame_t *st, size_t length)
 {
     unsigned int offset = 0;
+    unsigned int audio_end = length;
 
     if (has_fixed(st))
-        length = process_fixed_data(st, length);
+        audio_end = process_fixed_data(st, length);
 
-    while (offset < length - RS_CODEWORD_LEN)
+    while (offset < audio_end - RS_CODEWORD_LEN)
     {
         unsigned int start = offset;
         unsigned int j, lc_bits, loc_bytes, prog;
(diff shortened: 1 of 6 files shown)


src/frame.c
```diff
@@ -472,7 +472,11 @@ void frame_process(frame_t *st, size_t length)
         hef_t hef = {0};
 
         if (!fix_header(st, st->buffer + offset))
+        {
+            if (offset == 0)
+                st->input->sync_state = SYNC_STATE_NONE;
```
awesie: Do you want to always do this or after N (e.g. 2 or 3) failures? I'm just imagining a case where there is some periodic interference or weak signal.

argilo: I'll do some experimenting with that. I've done some testing, and the current design can result in significantly fewer audio packets being decoded when signal conditions are marginal.

argilo: @awesie It turns out the difference in the number of audio packets being decoded was only because some stations are broadcasting P3 PDUs that don't seem to have valid audio packets in them. This was causing unnecessary sync resets. I've added a second commit which ensures that a reset only occurs when a P1 PDU doesn't have any decodable audio packets. After this change, there's essentially no difference in performance.

awesie: This looks okay. A quick nitpick: maybe add a comment or something so that it is obvious that "length == MAX_PDU_LEN" is asserting that we are processing a P1 frame.

argilo: Done.

PR #175, opened 2019-02-10
Create a Python API and improve the Python CLI app

argilo:
To make Python development easier, I created a Python API (which wraps the C API), then rewrote the Python CLI app to use it. To exercise the API (both Python and C) I implemented almost all the features of the C CLI app, and added logging for SIG events as well. The Python CLI uses the same arguments as the C CLI.

This PR builds on top of #174 because the uninitialized XHDR values were preventing MIME values from being put into the `MIMEType` enum.

### DIFF
--- a/support/nrsc5.py
+++ b/support/nrsc5.py
@@ -1,61 +1,88 @@
-#!/usr/bin/env python3
-
 from ctypes import *
-import argparse
+import collections
 import enum
 import platform
-import pyaudio
-import queue
-import sys
-import threading
-import time
-
-audio_queue = queue.Queue(maxsize=32)
-hd_program = 0
-
-
-def audio_worker():
-    p = pyaudio.PyAudio()
-    try:
-        index = p.get_default_output_device_info()["index"]
-        stream = p.open(format=pyaudio.paInt16,
-                        channels=2,
-                        rate=44100,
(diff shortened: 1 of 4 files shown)


argilo:
Travis is happy: https://travis-ci.org/argilo/nrsc5/builds/491388057

support/nrsc5.py
```diff
@@ -1,153 +1,353 @@
...
 
-    libnrsc5.nrsc5_stop(radio)
-    libnrsc5.nrsc5_close(radio)
+        self.callback_func = CFUNCTYPE(None, POINTER(_Event), c_void_p)(callback_closure)
+        NRSC5.libnrsc5.nrsc5_set_callback(self.radio, self.callback_func, None)
 
-    audio_queue.put(None)
-    audio_thread.join()
+    def pipe_samples(self, samples):
+        result = NRSC5.libnrsc5.nrsc5_pipe_samples(self.radio, samples, (len(samples) // 4) * 4)
```
awesie: For completeness, is it worth adding a warning or assertion if len(samples) % 4 != 0?

argilo: Yeah, it's probably worth making it harder to use the API incorrectly. I'll add something in another PR.

PR #178, opened 2019-02-11
Get nrsc5 version through the API.

argilo:
It seems like querying the version of the nrsc5 library could be a useful thing, so I pushed the version number checking over to the other side of the API and made an `nrsc5_get_version` API call.

The Python CLI now has a `-v` flag, like the C version.

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -175,7 +175,7 @@ static void nrsc5_init(nrsc5_t *st)
     pthread_create(&st->worker, NULL, worker_thread, st);
 }
 
-NRSC5_API void nrsc5_get_version(char **version)
+NRSC5_API void nrsc5_get_version(const char **version)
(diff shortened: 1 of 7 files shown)


argilo:
Oops, the `-v` option works great in the Python CLI but nothing else does. :smile: Fixing now.

src/nrsc5.c
```diff
@@ -175,6 +175,11 @@ static void nrsc5_init(nrsc5_t *st)
     pthread_create(&st->worker, NULL, worker_thread, st);
 }
 
+NRSC5_API void nrsc5_get_version(char **version)
```
awesie: I would suggest using "const char **" to emphasize that the returned C string should not be modified nor freed.

argilo: Done.

PR #189, opened 2019-02-27
Use a matched filter for coarse sync.

argilo:
During coarse synchronization, `acquire_process` uses a rectangular (in the time domain) filter to detect the OFDM cyclic suffix. This is sub-optimal because the transmitter's pulse-shaping filter gives the beginning & end of the OFDM symbol a non-rectangular shape. To correct this, I've switched to a matched filter by multiplying in the beginning & end of the pulse-shaping filter.

To test, I forced coarse synchronization to be used all the time and compared the performance before and after this change using my suite of IQ recordings. There was a slight improvement in both BER and number of frames decoded, and one weak station that was previously unreceivable now works (albeit with a high BER).

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -63,7 +63,6 @@ void acquire_process(acquire_t *st)
     float angle, angle_diff, angle_factor, max_mag = -1.0f;
     int samperr = 0;
     unsigned int i, j, keep;
-    unsigned int mink = 0, maxk = FFTCP;
 
     if (st->idx != FFTCP * (ACQUIRE_SYMBOLS + 1))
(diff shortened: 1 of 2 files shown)


argilo:
Along the way, I also cleaned up some cruft that had accumulated in `acquire_process`. The change to the filter shape is on line 102, and the rest is cleanup.

awesie:
I don't have any means to test the performance improvements currently. The change looks reasonable, so if it improves performance go for it.

pclov3r:
Seems to have made some improvements here. Thought i'd add that.

PR #192, opened 2019-03-02
Add SIS messages to the API.

### LINKED ISSUE #176: Add SIS messages to the API
The SIS messages (station name, slogan, etc.) decoded in pids.c ought to produce events in the API. Some of them are pretty obscure (e.g. exciter core version), so they probably don't all need to be part of the API.

argilo:
Fixes #176.

This pull request adds Station Information Service (SIS) messages to the nrsc5 API (both C and Python).

Each time a new SIS message is received, the entire struct containing all SIS messages received so far is reported. This is a bit noisy for the current C & Python CLI applications, but seems appropriate for GUI applications.

"Station Name - long format" and "Station Slogan" messages are both returned in the `slogan` field (with Station Slogan taking precedence) since the NRSC-5 spec says that the long station name is deprecated in favor of slogan. (Some stations send the same data in both messages.)

A few things are not included in the API because I didn't think they were useful enough:

* ALFN
* Priority field for Station Message
* Control Data for Emergency Alert Message
* SIS Parameter Messages (time zone & leap second data, exciter & importer version numbers)

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -90,8 +90,28 @@ enum
     NRSC5_EVENT_AUDIO,
     NRSC5_EVENT_ID3,
     NRSC5_EVENT_SIG,
-    NRSC5_EVENT_LOT
+    NRSC5_EVENT_LOT,
+    NRSC5_EVENT_SIS
 };
+
+struct nrsc5_sis_asd_t
+{
+    struct nrsc5_sis_asd_t *next;
+    unsigned int program;
+    unsigned int access;
+    unsigned int type;
+    unsigned int sound_exp;
+};
+typedef struct nrsc5_sis_asd_t nrsc5_sis_asd_t;
+
+struct nrsc5_sis_dsd_t
+{
+    struct nrsc5_sis_dsd_t *next;
+    unsigned int access;
+    unsigned int type;
+    uint32_t mime_type;
+};
+typedef struct nrsc5_sis_dsd_t nrsc5_sis_dsd_t;
 
 struct nrsc5_event_t
 {
@@ -146,6 +166,19 @@ struct nrsc5_event_t
         struct {
             nrsc5_sig_service_t *services;
         } sig;
+        struct {
+            const char *country_code;
+            int fcc_facility_id;
+            const char *name;
+            const char *slogan;
+            const char *message;
+            const char *alert;
+            float latitude;
+            float longitude;
+            int altitude;
+            nrsc5_sis_asd_t *audio_services;
+            nrsc5_sis_dsd_t *data_services;
+        } sis;
     };
 };
 typedef struct nrsc5_event_t nrsc5_event_t;
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -25,6 +25,11 @@
 #define NRSC5_MIME_TEXT             0xBB492AAC
 #define NRSC5_MIME_JPEG             0x1E653E9C
 #define NRSC5_MIME_PNG              0x4F328CA0
+#define TTN_TPEG_1                  0xB39EBEB2
+#define TTN_TPEG_2                  0x4EB03469
+#define TTN_TPEG_3                  0x52103469
+#define TTN_STM_TRAFFIC             0xFF8422D7
+#define TTN_STM_WEATHER             0xEF042E96
 
 /*
  * Data types.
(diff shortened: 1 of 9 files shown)


src/pids.c
```diff
@@ -96,6 +97,99 @@ char *utf8_encode(int encoding, char *buf, int len)
     return NULL;
 }
 
+void report(pids_t *st)
+{
+    int i;
+    char *country_code = NULL;
+    char *name = NULL;
```
awesie: Since ```country_code``` and ```name``` should not be freed, consider marking them as ```const char *```.

src/pids.c
```diff
@@ -96,6 +97,99 @@ char *utf8_encode(int encoding, char *buf, int len)
...
+            dsd->type = st->data_services[i].type;
+            dsd->mime_type = st->data_services[i].mime_type;
+            data_services = dsd;
+        }
+    }
+
+    nrsc5_report_sis(st->input->radio, country_code, st->fcc_facility_id, name, slogan, message, alert,
+                     latitude, longitude, altitude, audio_services, data_services);
+
+    if (st->slogan_displayed)
```
awesie: I think it would be generally better if you could just always call ```free``` on these pointers. I understand why you can't because of slogan is sometimes a pointer that needs freed and sometimes it is a pointer that should not be freed.

Maybe you should have another variable (which is ```const char *``` and never freed) that contains either long name or slogan. That way you can just always free these pointers. I'm not committed to this, just don't like the idea of a pointer sometimes being freeable and sometimes not freeable.

src/pids.c
```diff
@@ -96,6 +97,99 @@ char *utf8_encode(int encoding, char *buf, int len)
     return NULL;
 }
 
+void report(pids_t *st)
```
awesie: ```static void ...```

src/main.c
```diff
@@ -291,6 +294,39 @@ static void callback(const nrsc5_event_t *evt, void *opaque)
...
+        if (evt->sis.slogan)
+            log_info("Slogan: %s", evt->sis.slogan);
+        if (evt->sis.message)
+            log_info("Message: %s", evt->sis.message);
+        if (evt->sis.alert)
+            log_info("Alert: %s", evt->sis.alert);
+        if (!isnan(evt->sis.latitude))
+            log_info("Station location: %f, %f, %dm", evt->sis.latitude, evt->sis.longitude, evt->sis.altitude);
+
+        audio_service = evt->sis.audio_services;
```
awesie: You could use a for-loop instead of the while-loops if you want.

include/nrsc5.h
```diff
@@ -25,6 +25,11 @@
 #define NRSC5_MIME_TEXT             0xBB492AAC
 #define NRSC5_MIME_JPEG             0x1E653E9C
 #define NRSC5_MIME_PNG              0x4F328CA0
+#define TTN_TPEG_1                  0xB39EBEB2
```
awesie: Should we put a ```NRSC5_``` prefix on these for namespacing?

src/pids.c
```diff
@@ -135,9 +229,9 @@ void decode_sis(pids_t *st, uint8_t *bits)
 
             if ((strcmp(country_code, st->country_code) != 0) || (fcc_facility_id != st->fcc_facility_id))
             {
-                log_debug("Country: %s, FCC facility ID: %d", country_code, fcc_facility_id);
                 strcpy(st->country_code, country_code);
                 st->fcc_facility_id = fcc_facility_id;
+                report(st);
```
awesie: It looks like ```report()``` is going to potentially be called many times as the structure is updated. Might it be better to have a ```changed``` variable and call ```report()``` once at the end of the function if ```changed == true```?

include/nrsc5.h
argilo: Yes; I forgot to do that when i copied them over from Python.

src/pids.c
argilo: I might try that out, but the gains won't be huge since there are at most two SIS messages per PIDS frame.

src/pids.c
argilo: Done.

src/pids.c
argilo: I added a `strdup` to the `long_message` case, and made the `free`s below unconditional.

src/pids.c
argilo: The improvement turned out to be about 40%. I also added a couple extra checks to make sure that location is only reported when both latitude and longitude are ready. (They arrive in two separate messages.)

argilo:
@awesie I made all the improvements you suggested, so this is ready for another look now.

PR #194, opened 2019-03-02
Add signed 16-bit sample input.

argilo:
The `nrsc5_pipe_samples` function currently takes unsigned 8-bit samples at a sample rate of 1488375 Hz. This is convenient for the RTL-SDR but not for other SDRs, many of which have an ADC resolution higher than 8 bits.

To solve this problem, I've split `nrsc5_pipe_samples` into two functions:

* `nrsc5_pipe_samples_cu8` which uses the current RTL-SDR format
* `nrsc5_pipe_samples_cu16` which accepts 16-bit signed samples at a sample rate of 744187.5 Hz

The new `nrsc5_pipe_samples_cu16` function puts samples straight into the input module's buffer bypassing the halfband filter / decimator. 

To test the new functions, I've added an `--iq-input-format` argument to the Python CLI which defaults to the old RTL-SDR format. I received samples into a FIFO using a USRP B200:
```
mkfifo samples
uhd_rx_cfile -f 100.3e6 -r 744188 -g 50 -A TX/RX -s samples
```
And in a separate terminal I ran `cli.py`:
```
cli.py --iq-input-format cs16 -r samples 0
```

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -59,9 +59,9 @@ static inline cint16_t cf_to_cq15(float complex x)
     return cq15;
 }
 
-static inline float complex cq15_to_cf(cint16_t cq15)
+static inline float complex cq15_to_cf_conj(cint16_t cq15)
 {
-    return CMPLXF((float)cq15.r / 32767.0f, (float)cq15.i / 32767.0f);
+    return CMPLXF((float)cq15.r / 32767.0f, (float)cq15.i / -32767.0f);
 }
 
 static inline float normf(float complex v)
--- a/support/nrsc5.py
+++ b/support/nrsc5.py
@@ -352,9 +352,16 @@ class NRSC5:
         self.callback_func = CFUNCTYPE(None, POINTER(_Event), c_void_p)(callback_closure)
         NRSC5.libnrsc5.nrsc5_set_callback(self.radio, self.callback_func, None)
 
-    def pipe_samples(self, samples):
+    def pipe_samples_cu8(self, samples):
         if len(samples) % 4 != 0:
             raise NRSC5Error("len(samples) must be a multiple of 4.")
-        result = NRSC5.libnrsc5.nrsc5_pipe_samples(self.radio, samples, len(samples))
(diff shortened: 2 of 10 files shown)


src/defines.h
```diff
@@ -59,9 +59,9 @@ static inline cint16_t cf_to_cq15(float complex x)
     return cq15;
 }
 
-static inline float complex cq15_to_cf(cint16_t cq15)
+static inline float complex cq15_to_cf_conj(cint16_t cq15)
 {
-    return CMPLXF((float)cq15.r / 32767.0f, (float)cq15.i / 32767.0f);
+    return CMPLXF((float)cq15.r / 32767.0f, (float)cq15.i / -32767.0f);
```
argilo: It was more convenient to perform the complex conjugation here instead of in `input_push`.

support/nrsc5.py
```diff
@@ -352,9 +352,16 @@ def callback_closure(evt, opaque):
...
-    def pipe_samples(self, samples):
+    def pipe_samples_cu8(self, samples):
         if len(samples) % 4 != 0:
             raise NRSC5Error("len(samples) must be a multiple of 4.")
-        result = NRSC5.libnrsc5.nrsc5_pipe_samples(self.radio, samples, len(samples))
+        result = NRSC5.libnrsc5.nrsc5_pipe_samples_cu8(self.radio, samples, len(samples))
+        if result != 0:
+            raise NRSC5Error("Failed to pipe samples.")
+
+    def pipe_samples_cs16(self, samples):
```
awesie: This API is a little strange because in C samples is ```int16_t *``` but in Python it is a byte array.

argilo: I went that route because the most likely use case is reading in samples from a file, which will arrive in a `bytes` object. I wanted to avoid having to unpack & repack the data. If you have a better idea, I'd be happy to reconsider it.

awesie: I didn't have any immediate idea without relying on numpy, and I don't think it is so important that adding a dependency is worth it.

PR #195, opened 2019-03-03
Fix reporting of synchronization state.

argilo:
I noticed a couple problems with the reporting of synchronization state. First, loss of synchronization is not reported if it occurs in `frame_process`. Second, the synchronization state is logged in sync.c instead of main.c.

To fix these problems, I've created a new `input_set_sync_state` function which is called whenever the sync state needs to be updated. It reports sync events through the API where appropriate.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -114,7 +114,7 @@ void acquire_process(acquire_t *st)
         angle_factor = (st->prev_angle) ? 0.25 : 1.0;
         angle = st->prev_angle + (angle_diff * angle_factor);
         st->prev_angle = angle;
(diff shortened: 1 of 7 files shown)


argilo:
> Do you think we should report lost sync in input_reset as well?

Yes, I think that's a good idea. To avoid duplicate reports, I updated `input_set_sync_state` to check whether the new state differs from the old. `input_init` now initializes the sync state to NONE, and `input_reset` updates the state using `input_set_sync_state`.

This is ready for a final review.

argilo:
Rebased to resolve a merge conflict. I'll merge this once Travis is happy.

PR #196, opened 2019-03-03
Display XHDR even when param is 0.

argilo:
This fixes a small bug: the XHDR field was not displayed if `param` was zero. (This occurs when album art is displayed.)


argilo:
On second thought, I should probably just set the entire `xhdr` field to `None` in nrsc5.py when XHDR is absent. Then I can just check `if evt.xhdr:` in the CLI. I'll make that change.

argilo:
Done. I also made the same improvement to the API for `ufid`.

Issue #73, opened 2017-09-24
Switch program streams on the fly


pclov3r:
So i'm curious if it's possible to switch between the different programs a station may have without having quit the program in again and going though the entire process again.

I think this also goes back to one of my earlier issues in which you said this needs a GUI. 

Perhaps this is too much work or not possible. 

What do you think?

argilo:
It's definitely a feature worth having, so let's leave this issue open. Even without a GUI, it should be possible to accept keyboard input to switch streams. It would also be nice for nrsc5 to tell you which programs are present.

pclov3r:
Perhaps something among the lines of something line of a 0-3 numeric input would switch the stream so you wouldn't have to go tough the entire process having to re-acquire the channel again etc.

KeyserSoze1:
While mine was strictly just hacking something together for a demo this should be very easily done.  I was able to send a udp packet to nrsc5 to have it switch on the fly.  In my case I had gqrx send a udp packet to a specific port on localhost that nrsc5 was listening to.  There were no issues with this and it worked as expected.  If I have time I'll look at adding keyboard input for switching programs and push a pull request if @argillo doesn't get to it first.

pclov3r:
Figure you haven't had a chance to look into this. Still would like this feature :)

argilo:
This feature seems like it would be more appropriate for a GUI application where there's an easy way to accept user input. The nrsc5 API already supports this use case, since it reports data pertaining to all streams through the callback function. It's up to the application to decide which stream (or streams) it cares about.

Since we now have a Python API, I was thinking I might try switching @cmnybo's [nrsc5-gui](https://github.com/cmnybo/nrsc5-gui) to use it. Adding a feature to switch streams should be straightforward.

awesie:
I generally agree with @argilo. I put together a quick hack if we want to support it in the CLI, though I am fine with the answer being no.

argilo:
It's now possible to switch streams during playback in NRSC5-GUI as well. My changes are available in this pull request: https://github.com/cmnybo/nrsc5-gui/pull/9

PR #197, opened 2019-03-03
Add simple keyboard commands to CLI.

awesie:
Proof-of-concept for adding support for some basic commands like quit or change audio program. I only spent a few minutes on it, mainly to get something out there for feedback on resolving #73.

One issue is that the CLI doesn't keep track of valid program numbers to give user feedback if they select a missing program, though the recent SIS event support should make this possible.

Second issue is that there is still a delay because we don't buffer audio from programs that are not currently selected.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -414,7 +414,8 @@ static void *input_main(void *arg)
 #ifdef __MINGW32__
         ch = _getch();
 #else
-        ch = getchar();
+        if (read(STDIN_FILENO, &ch, 1) != 1)
+            break;
 #endif
 
         switch (ch)
--- a/src/main.c
+++ b/src/main.c
@@ -700,6 +700,7 @@ int main(int argc, char *argv[])
     pthread_cancel(input_thread);
     pthread_join(input_thread, NULL);
 
+    nrsc5_stop(radio);
     nrsc5_close(radio);
     cleanup(st);
     free(st);
--- a/src/main.c
+++ b/src/main.c
@@ -19,6 +19,7 @@
 #include <nrsc5.h>
 #include <pthread.h>
 #include <sys/time.h>
+#include <unistd.h>
 
 #include <assert.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


pclov3r:
I get this when attempting to compile with this PR.

main.c:387:15: error: ‘STDIN_FILENO’ undeclared (first use in this function)
     tcsetattr(STDIN_FILENO, TCSANOW, arg);

including unistd.h resolved it. 

As for Windows termios.h does not exist in mingw-w64 and this is likely to cause some problems as windows isn't POSIX compliant.

pclov3r:
For windows I found this https://github.com/chjchoi/kai/blob/master/project/linux-cortexm-1.9.0/A2F/gdb-2011.03/cs-cygwin-wrapper/mingw-termios.h which seems to resolve most of it.

However, It's still telling me tcsetattr is undefined but it's defined here. Hum...

I'm not a programmer so somebody else will have to pitch in here.

awesie:
I fixed the build errors on Windows and Travis CI.

pclov3r:
Thanks! it seems to work now.

argilo:
Is Q meant to exit the program? It only stops the audio for me (on Linux).

awesie:
Thanks. I think this may be because I only tested with a file input.

awesie:
I think I fixed it by adding a call to ```nrsc5_stop``` before ```nrsc5_close```. Though it might make more sense for nsc5_close to make sure that the worker is stopped itself.

src/main.c
```diff
@@ -347,6 +385,66 @@ static void callback(const nrsc5_event_t *evt, void *opaque)
...
 }
 
+#ifndef __MINGW32__
+static void restore_termios(void *arg)
+{
+    tcsetattr(STDIN_FILENO, TCSANOW, arg);
+}
+#endif
+
+static void *input_main(void *arg)
```
argilo: Make `arg` a `state_t *`?

argilo: Never mind, it's `void *` because it's invoked by pthreads.

awesie: Yeah, I considered assigning it to a local variable just to make it explicit that it should be a ```state_t *``` rather than rely on the implicit conversion.

PR #200, opened 2019-03-09
Optimize a few things.

argilo:
I profiled nrsc5 and made some optimizations to the parts that use the most CPU cycles. The result is a speedup of about 5%.

The most significant changes are in sync.c where I updated some loops to process only the FFT bins which might later be used (i.e. the ones that may contain digital subcarriers).

### DIFF
--- a/src/sync.c
+++ b/src/sync.c
@@ -23,6 +23,7 @@
 #include "sync.h"
 
 #define PM_PARTITIONS 10
+#define MAX_PARTITIONS 14
 #define PARTITION_DATA_CARRIERS 18
 #define PARTITION_WIDTH 19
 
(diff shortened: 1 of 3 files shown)


src/sync.c
```diff
@@ -23,6 +23,7 @@
 #include "sync.h"
 
 #define PM_PARTITIONS 10
+#define MAX_PARTITIONS 14
```
awesie: I assume that this number came from the for-loop on line 221 (PM_PARTITIONS +/- 2)? If so, maybe add a define to make that relationship clearer. Something like:
```
#define CFO_PARTITIONS 2
#define MAX_PARTITIONS (PM_PARTITIONS + CFO_PARTITIONS)
//...
for (i = -CFO_PARTITIONS * PARTITION_WIDTH; i < CFO_PARTITIONS * PARTITION_WIDTH; ++i)
```

argilo: It's actually the maximum of these `partitions_per_band` numbers:

https://github.com/theori-io/nrsc5/blob/a40277e2686ea59a1a61d573b521c7adf8c9bfb5/src/sync.c#L173-L187

PR #201, opened 2019-03-10
Speed up Viterbi decoding

argilo:
NRSC-5 uses convolutional codes in tail-biting mode. Currently the convolutional decoder in nrsc5 performs two full passes of Viterbi decoding. This improves decoding, but doubles the computational cost. Extending the beginning & end of the trellis by 32 achieves the same decoding performance at much lower computational cost. Making this change improves the overall performance of nrsc5 by about 7%.

In this PR I've also removed the unused functions in conv_neon.h and conv_sse.h to silence compiler warnings.

### DIFF
--- a/src/conv_dec.c
+++ b/src/conv_dec.c
@@ -41,6 +41,7 @@
 #endif
 
 #define PARITY(X) __builtin_parity(X)
+#define TAIL_BITING_EXTRA 32
 
 /*
  * Trellis State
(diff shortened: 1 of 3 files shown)


argilo:
To test the decoding performance I used my usual suite of ~50 I/Q recordings and looked at the average BER. The numbers were identical in all but two cases, and the final two were worse by 0.000001 and 0.000002 respectively.

PR #208, opened 2019-07-28
Fix WAV file output on Windows

### LINKED ISSUE #207: [Windows] File output very noisy
Directing output to a file results in very noisy audio file with lots of hiss and scratching.  The problem is present with both MSYS-32 and MSYS-64 builds.

argilo:
Fixes #207.

When opening a WAV file for writing, libao doesn't request binary mode. On Windows, this causes carriage return characters to be inserted into the output file before line feed characters, corrupting the file.

For now I've modified nrsc5 to patch libao's `ao_open_file` function. I'll propose the patch to libao and remove it from here if the patch is accepted upstream.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -189,6 +189,7 @@ if (NOT USE_SYSTEM_LIBAO)
         PREFIX ${LIBAO_PREFIX}
 
         UPDATE_COMMAND ""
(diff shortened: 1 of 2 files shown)


awesie:
LGTM

argilo:
I submitted a merge request to libao: https://gitlab.xiph.org/xiph/libao/merge_requests/1

PR #214, opened 2020-01-24
Fix undefined behavior

argilo:
When nrsc5 is built with `-fsanitize=undefined` to detect undefined behavior, some errors appear:
```
/home/argilo/git/nrsc5/src/input.c:173:18: runtime error: left shift of negative value -9
/home/argilo/git/nrsc5/src/input.c:171:18: runtime error: left shift of negative value -57
/home/argilo/git/nrsc5/src/input.c:172:18: runtime error: left shift of negative value -33
/home/argilo/git/nrsc5/src/input.c:174:18: runtime error: left shift of negative value -53
/home/argilo/git/nrsc5/src/output.c:401:79: runtime error: left shift of 190 by 24 places cannot be represented in type 'int'
/home/argilo/git/nrsc5/src/output.c:613:80: runtime error: left shift of 187 by 24 places cannot be represented in type 'int'
```
I made the necessary changes to avoid undefined behaviour and verified that the errors went away.

I also changed the types of the `lot` and `size` fields in the `aas_file_t` struct to be consistent with the packet format.

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -45,7 +45,7 @@
 #define MAX_PROGRAMS 8
 
 #define U8_F(x) ( (((float)(x)) - 127) / 128 )
-#define U8_Q15(x) ( ((int16_t)(x) - 127) << 6 )
+#define U8_Q15(x) ( ((int16_t)(x) - 127) * 64 )
 
 typedef struct {
     int16_t r, i;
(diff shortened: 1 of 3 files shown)


argilo:
Rebased to pick up 9751a6183154a0ed0202b2be12bbcfe852ad70fb, which was needed because wine was removed from Homebrew Core.

Hopefully the Travis build will succeed now: https://travis-ci.org/argilo/nrsc5/builds/641213233

awesie:
lgtm

PR #222, opened 2020-06-14
Remove libao patch

argilo:
The libao patch I added in #208 has now been accepted upstream:

https://gitlab.xiph.org/xiph/libao/-/merge_requests/1

As a result, we can remove the patch from nrsc5. The GitHub mirror of libao is not up to date, so I've switched to gitlab.xiph.org for now.


argilo:
Tested on Windows and WAV output is still working properly. Merging.

PR #223, opened 2020-06-20
Consider all reference subcarriers when detecting CFO

### LINKED ISSUE #221: Single iBOC sideband decoding?
So it appears that if a sideband is completely blocked or has horrible multipath NRSC5 will not decode any audio. Other hardware radios seem to be able to lock and decode with only a a good single iBOC sideband present. 

Not sure what the NRSC-5 spec says about this or if it should or shouldn't be supported.

I've some some people claim that with a station that only has 2 HD streams both sidebands contain duplicate information. For HD3 and 4 stations both need to be present. Not sure how true this is. 

Thanks!

argilo:
Fixes #221.

As noted in https://github.com/theori-io/nrsc5/issues/221#issuecomment-643716708, CFO synchronization uses two specific reference subcarriers. If either of those subcarriers is missing (due to interference or multipath), then the entire synchronization process fails.

To fix this, I've changed the CFO synchronization routine to take into account all 22 reference subcarriers in the Primary Main (PM) sidebands. If at least three of them are found to have the same offset, then we assume that we've found the correct synchronization parameters.

### DIFF
--- a/src/sync.c
+++ b/src/sync.c
@@ -16,6 +16,7 @@
 #include "config.h"
 
 #include <math.h>
+#include <string.h>
 
 #include "defines.h"
 #include "input.h"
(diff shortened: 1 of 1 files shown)


argilo:
I tested this by filtering out the entire upper (or lower) sideband using a GNU Radio flowgraph. After the change, I can receive my local station (CJMJ) with either of the sidebands removed.

pclov3r:
Things are looking good here doing some testing.

argilo:
I ran this against my test suite of IQ recordings (which have both sidebands present) and the total amount of decoded audio packets increased slightly.

src/sync.c
```diff
@@ -220,23 +222,41 @@ void sync_process(sync_t *st)
         }
         else if (st->cfo_wait == 0)
         {
-            for (i = -2 * PARTITION_WIDTH; i < 2 * PARTITION_WIDTH; ++i)
+            for (cfo = -2 * PARTITION_WIDTH; cfo < 2 * PARTITION_WIDTH; cfo++)
```
awesie, line 225: I wouldn't be opposed to pulling this code out to a separate function if you want. It looks independent of the other code.

src/sync.c
```diff
@@ -220,23 +222,41 @@ void sync_process(sync_t *st)
...
+
+                for (offset = 0; offset < BLKSZ; offset++)
+                {
+                    if (offset_count[offset] > best_count) {
+                        best_offset = offset;
+                        best_count = offset_count[offset];
+                    }
+                }
+
+                if (best_offset >= 0 && offset_count[best_offset] >= 3)
```
awesie, line 253: Minor comment: you use `offset_count[best_offset]` in here but `best_count` above

argilo:
Thanks for the suggestions @awesie. I refactored accordingly, and verified that all my test files are still decoding properly.

PR #228, opened 2020-06-24
Basic rtl_tcp support

### LINKED ISSUE #86: rtl_tcp support
So I thought it might be useful to have rtl_tcp support built in.

awesie:
Fixes #86  

I quickly threw this together so there might be some tweaking we want to do on it. I tested the basic functionality with a Linux rtl_tcp server and nrsc5 clients on each OS.

Future improvements:

- optionally set ppm error (since it is likely already set on the server-side)
- auto-gain support (We know the number of gain levels, and we can iterate over them by index. We do not know the actual gain themselves though.)


argilo:
Nice! I took this for a quick spin just now and it works well. I don't have time for code review tonight, but I'll take a look tomorrow after work.

awesie:
How do we feel about removing `ppm_error` option from `nrsc5_open` and instead having a separate external API for it, similar to how we have it for frequency and gain?

argilo:
> How do we feel about removing ppm_error option from nrsc5_open and instead having a separate external API for it, similar to how we have it for frequency and gain?

I agree there's no good reason for it to be set up front in the `nrsc5_open` functions, so a separate `nrsc5_set_freq_correction` would be a better place.

That said, we could even consider removing frequency correction entirely. The CFO detection routine can tolerate a frequency error as large as 38 subcarriers:

https://github.com/theori-io/nrsc5/blob/8074b1c523a104df8ca325dc12a3e4b3af6a7cda/src/sync.c#L172

In the worst case (at the upper end of the FM band) that's ±128 ppm, and I believe the worst RTL-SDR dongles are ±120 ppm.

src/CMakeLists.txt
```diff
@@ -29,6 +30,7 @@ link_directories (
     ${AO_LIBRARY_DIRS}
     ${RTL_SDR_LIBRARY_DIRS}
     ${FFTW3F_LIBRARY_DIRS}
+    ${LIBUSB_LIBRARY_DIRS}
```
argilo, line 33: Should there be a corresponding entry in `include_directories` above?

src/rtltcp.c
```diff
@@ -0,0 +1,193 @@
...
+        if (err <= 0)
+            break;
+        recvd += err;
+    }
+#ifdef __MINGW32__
+    mode = 0;
+    ioctlsocket(st->socket, FIONBIO, &mode);
+#endif
+    // if we read an odd-number of bytes, read one more byte
+    recv(st->socket, buf, 1, 0);
```
argilo, line 178: Is this meant to be inside an `if`?

src/CMakeLists.txt
awesie: The include directory isn't really necessary because we don't use libusb ourselves. If we build librtlsdr then we need to be able to link to libusb.

src/rtltcp.c
awesie: Ah, yes. I'll fix that and make sure it still works.

PR #230, opened 2020-06-27
Improve synchronization

argilo:
I noticed that #223 made CFO detection worse when there is a frequency error of more than a few parts per million. The reason is that `adjust_ref` is now called multiple times for each subcarrier as different CFO values are tried, and so the symbol phases get adjusted multiple times. To fix this, I added a `reset_ref` function to undo the phase adjustments after each offset is tried.

I also fixed some other problems with the synchronization code:

* The CFO frequency calculation in `adjust_ref` was incorrect.
* The `sync_reset` function didn't reset the Costas loop parameter for some of the subcarriers that were accessed during CFO detection. I changed it to reset all subcarriers.
* Checking the validity of the reference subcarriers is unnecessary in the `SYNC_STATE_FINE` state, since `frame_process` already detects loss of sync. I removed this check.
* In the `SYNC_STATE_COARSE` state, one of two specific reference subcarriers must be present to trigger a transition into the `SYNC_STATE_FINE` state. If both are corrupted, then synchronization will not be achieved. I changed this to proceed if any four reference subcarriers are successfully decoded.

Overall these changes made an 0.3% improvement in the number of decoded audio packets in my suite of I/Q recordings. For high-SNR recordings there was no difference, but for low-SNR recordings the improvement was as high as 15%.

### DIFF
--- a/src/sync.c
+++ b/src/sync.c
@@ -31,7 +31,7 @@
 static void adjust_ref(sync_t *st, unsigned int ref, int cfo)
 {
     unsigned int n;
-    float cfo_freq = 2 * M_PI * cfo * CP / FFTCP;
+    float cfo_freq = 2 * M_PI * cfo * CP / FFT;
 
     // sync bits (after DBPSK)
     static const signed char sync[] = {
@@ -67,6 +67,12 @@ static void adjust_ref(sync_t *st, unsigned int ref, int cfo)
         }
         st->costas_phase[ref] += M_PI;
     }
+}
+
+static void reset_ref(sync_t *st, unsigned int ref)
+{
+    for (unsigned int n = 0; n < BLKSZ; n++)
+        st->buffer[ref][n] *= cexpf(I * st->phases[ref][n]);
 }
 
 static void decode_dbpsk(const float complex *buf, unsigned char *data, int size)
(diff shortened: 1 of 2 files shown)


src/sync.c
```diff
@@ -181,12 +187,14 @@ void detect_cfo(sync_t *st)
         for (int i = 0; i <= PM_PARTITIONS; i++)
         {
             adjust_ref(st, cfo + LB_START + i * PARTITION_WIDTH, cfo);
-            offset = find_ref(st, cfo + LB_START + i * PARTITION_WIDTH, (PM_PARTITIONS-i) & 0x3);
+            offset = find_ref(st, cfo + LB_START + i * PARTITION_WIDTH, (30-i) & 0x3);
```
awesie, line 190: In a few places you use the constant `30`, what is this referring to?

argilo: This comes from table 11-3 (Reference Subcarrier Identification) in 1011s.pdf. The RSID bits are symmetric about column 30. So column 30 has an RSID of 0, columns 29 and 31 have an RSID of 1, column 28 and 32 have an RSID of 2, etc. I struggled to come up with a good name for this, but I'll think of something.

argilo:
> Should that instead be stored in `sync_t`

Yes, I think that's a better place for it.

Thanks for all the suggestions. I'll make those changes later tonight.

src/sync.c
awesie: Ah ok. It doesn't need to be a good name, just preferably something where you can add a comment like that.

PR #231, opened 2020-07-08
Added function to set the log level to nrsc5 library.

evanburnsdev:
Discovered I was unable to set the log level to suppress all the LOG_DEBUG calls generated by the library.  I thought this would be handy for anyone else using the library with their code.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -257,6 +257,7 @@ typedef struct nrsc5_t nrsc5_t;
 /*
  * Public functions. All functions return void or an error code (0 == success).
  */
+void nrsc5_log_set_level(int level);
 void nrsc5_get_version(const char **version);
 void nrsc5_service_data_type_name(unsigned int type, const char **name);
 void nrsc5_program_type_name(unsigned int type, const char **name);
--- a/src/libnrsc5.map
+++ b/src/libnrsc5.map
@@ -1,5 +1,6 @@
 LIBNRSC5_1.0 {
     global:
+        nrsc5_log_set_level;
         nrsc5_get_version;
         nrsc5_service_data_type_name;
         nrsc5_program_type_name;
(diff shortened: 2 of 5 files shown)


pclov3r:
@argilo Any opinions on this? :)

argilo:
Back when @awesie split nrsc5 into a library and a command-line app, I believe the plan was to remove logging from the library completely. Maybe we should do that instead.

I suspect a lot of the log messages were only relevant during development, and could now be removed. I'll have look to see what can go.

evanburnsdev:
Awesome!  Thanks for the update.

argilo:
I opened #244 to replace this. Let me know what you think, @evanburnsdev.

evanburnsdev:
#244 will definitely do what I need.  The only case I could see it being an issue for someone would be if a library user needed to troubleshoot their code, they'd have to recompile the library every time they wanted to change the debug level.
I could be reading it wrong, I'm looking at the code from my phone.  Ill have a chance to sit down and test it this weekend if thats not too far away for you.

argilo:
That's correct; turning on debug logging would require recompiling the library. I did it that way because I didn't think the debug messages would be very useful to users of the library. They're aimed at people developing the library itself.

argilo:
If any of them *are* useful to users of the library, then I think we should expose them through the callback mechanism.

evanburnsdev:
I did use them at first to help me understand how to use the library functions properly.  In my case, it was nice to see when the album/station artwork was downloaded, so I could narrow down why it wasn't being displayed properly for me(which of course is because I had an error in my program).  I'm sure the other metadata being displayed could be helpful to others in similar situations.
Either way, recompiling isn't the end of the world and your solution is perfect for me at this point.  Just trying to think about what others may run into.

argilo:
Thanks for the feedback.

I updated #244 so that it's easier to re-enable logging (by adding `-DLIBRARY_DEBUG_LEVEL=1` to the cmake command). Since that pull request has superseded this one, I'll close this off now.

evanburnsdev:
Thank you!

PR #232, opened 2020-08-04
Add support for AM

### LINKED ISSUE #33: AM NRSC-5
Would it be possible to add support for the AM Broadcast band? From what I understand the sidebands are much more narrow and use much higher order modulation.

argilo:
Resolves #33.

This adds basic support for receiving AM signals, as described in [NRSC-5 Layer 1 AM](https://www.nrscstandards.org/standards-and-guidelines/documents/standards/nrsc-5-d/reference-docs/1012s.pdf).

To use the new functionality, simply add the `--am` command line option.

Some limitations:

* Only the hybrid mode (MA1) is supported. Fortunately, almost all stations (except WWFD) are using this mode.
* Audio quality is sub-optimal, because only the "core" audio stream (20 kbps) is decoded. I don't yet know how to decode the packets in the "enhanced" audio stream, which contains an additional 16 kbps.
* Receive performance is sub-optimal. In particular, decimation and OFDM synchronization could be improved, and the Viterbi decoder could be switched from hard to soft decisions.
* Automatic gain selection is not implemented. The `-g` option should be used to manually specify a gain value.

Since the code is in a working state and the number of changes is already large, I thought it would make sense to merge in the feature now and continue working on it in subsequent pull requests.

I used my test suite of I/Q recordings to verify that the FM receiver functions identically before & after these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ Replace `32` with `64` if you want a 64-bit build. Once the build is complete, c
        -l log-level                    set log level
                                          (1 = DEBUG, 2 = INFO, 3 = WARN)
        -v                              print the version number and exit
+       --am                            receive AM signals
+                                         (default is FM)
        --dump-aas-files dir-name       dump AAS files
                                          (WARNING: insecure)
        --dump-hdc file-name            dump HDC packets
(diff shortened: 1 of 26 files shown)


pclov3r:
@awesie @argilo Any progress on the review to maybe approve this PR? :)

argilo:
I expect @awesie has been very busy with the DEF CON CTF this past week.

The code here is working and doesn't appear to break anything for FM, so I'll just merge it for now. I'll be happy to incorporate any suggested improvements in a follow-up PR.

pclov3r:
@argilo Thanks! And thanks for your work on this.

PR #242, opened 2020-08-16
Add support for bias-T & direct sampling

### LINKED ISSUE #235: Add option to turn on Bias-T
With the addition of AM support it would be useful to be able to enable the Bias-t on RTL-SDR's that have this supported such as the RTL-SDR Blog V3 unit. You could use it to power say an an active loop antenna such as the popular MLA-30.

### LINKED ISSUE #236: Add direct sampling support for AM
While the rtl-sdr does not natively support tuning to the AM broadcast band you can do so with a modified unit or one that is already enabled to do direct sampling mode such as the RTl-SDR Blog unit V3.

argilo:
Fixes #235.
Fixes #236.

`-T` enables the bias-T.
`-D 1` enables direct sampling on the I-ADC input
`-D 2` enables direct sampling on the Q-ADC input (which works with RTL-SDR Blog V3 dongles)

All the standard `rtl_*` utilities use `-T` for the bias-T, so that seemed like a good choice. The `rtl_power` tool uses `-D` for direct sampling, although without an argument to choose the input.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 
-dist: xenial
+dist: bionic
 
 compiler:
   - gcc
@@ -18,7 +18,6 @@ before_install:
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -y libao-dev libfftw3-dev librtlsdr-dev mingw-w64 python3-pyaudio; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo ln -s /usr/bin/ccache /usr/lib/ccache/i686-w64-mingw32-gcc; sudo ln -s /usr/bin/ccache /usr/lib/ccache/x86_64-w64-mingw32-gcc; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew cask install wine-stable; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install libao fftw librtlsdr mingw-w64 portaudio ccache; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export PATH="/usr/local/opt/ccache/libexec:$PATH"; ln -s /usr/local/bin/ccache /usr/local/opt/ccache/libexec/i686-w64-mingw32-gcc; ln -s /usr/local/bin/ccache /usr/local/opt/ccache/libexec/x86_64-w64-mingw32-gcc; fi
--- a/README.md
+++ b/README.md
@@ -103,6 +103,9 @@ Replace `32` with `64` if you want a 64-bit build. Once the build is complete, c
        -v                              print the version number and exit
        --am                            receive AM signals
                                          (default is FM)
+       -T                              enable bias-T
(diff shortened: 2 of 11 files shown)


argilo:
@pclov3r Could you help test this?

argilo:
The new options should work in rtl_tcp mode also.

argilo:
Unfortunately there's a snag: the `rtlsdr_set_bias_tee` function was only added in rtl-sdr version 0.5.4, released in April 2018. Ubuntu 16.04 and 18.04 both ship with 0.5.3, and so nrsc5 fails to build there. I guess we'll need to detect whether or not the function is present and disable the functionality if necessary.

argilo:
Actually Ubuntu 18.04 should be fine, since bias-T support was pulled in before the 0.5.4 release of rtl-sdr: http://changelogs.ubuntu.com/changelogs/pool/universe/r/rtl-sdr/rtl-sdr_0.5.3-13/changelog

In that case, maybe it's not so bad to depend on the function being present. I'll try switching the Travis build to Ubuntu 18.04 and see how it goes.

.travis.yml
```diff
@@ -18,7 +18,6 @@ before_install:
 
 install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get install -y libao-dev libfftw3-dev librtlsdr-dev mingw-w64 python3-pyaudio; fi
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo ln -s /usr/bin/ccache /usr/lib/ccache/i686-w64-mingw32-gcc; sudo ln -s /usr/bin/ccache /usr/lib/ccache/x86_64-w64-mingw32-gcc; fi
```
argilo, line 21: It looks like this is no longer necessary on Ubuntu 18.04:
```
ln: failed to create symbolic link '/usr/lib/ccache/i686-w64-mingw32-gcc': File exists
ln: failed to create symbolic link '/usr/lib/ccache/x86_64-w64-mingw32-gcc': File exists
```
My second build was much faster than the first, so ccache appears to be working with mingw.

argilo:
All is well after updating Travis-CI to use Ubuntu 18.04: https://travis-ci.org/github/argilo/nrsc5/builds/718395860

pclov3r:
@argilo I'll see what I can do about testing this over the next few days. Been a bit busy of recent.

With that said trying to get a decode of the one station i see might be a bit difficult. With that said I can test the Bias-T and Direct sampling mode tho and that it's functional.

argilo:
> With that said it seems quitting on windows at least does not disable the Bias-T.

The other `rtl_*` tools behave similarly, so I didn't make any effort to turn off the bias-T on exit. It probably would be friendlier to attempt to turn it off though, so I'll see what I can do about that.

argilo:
Updated: the bias-T now turns off if Q is pressed in nrsc5, or if Ctrl+C is pressed in cli.py.

Ctrl+C doesn't turn it off in nrcs5 at the moment because it hasn't got a SIGINT handler, and so the cleanup code doesn't have a chance to execute. That's probably something for another pull request, if we decide to change it.

awesie (approved):
LGTM, but I don't have equipment for testing this.

PR #244, opened 2020-08-20
Remove logging from the library by default

argilo:
Replaces #231.

I've replaced the nrsc5 library's logging with simple macros that do nothing unless `DEBUG_LEVEL` is changed to something less than 5.


awesie:
My original thinking when I split out the library was to eventually do something libusb (API user can register a callback function that receives the log messages): http://libusb.sourceforge.net/api-1.0/group__libusb__lib.html#ga2efb66b8f16ffb0851f3907794c06e20.

I assume this pull request replaces all calls to the `log_` functions? If so, it should probably remove log.c and log.h, and the log level should be a compile-time option with a default (instead of hard-coded in a header file). If we decide later we need more logging, we can always add an API option later.

argilo:
Thanks for weighing in on this @awesie.

> I assume this pull request replaces all calls to the `log_` functions? If so, it should probably remove log.c and log.h 

I've only removed them from the library portion; main.c still uses the logging from log.c and log.h.

> the log level should be a compile-time option with a default (instead of hard-coded in a header file)

Agreed, I'll update the pull request to do it that way.

argilo:
I've made the log level a compile-time option as suggested. I don't work with C or cmake that much, so please let me know if there's a better way to do this.

awesie:
It is probably okay for now. It is similar to how you added `GIT_COMMIT_HASH`.

The alternative way is to use the config.h.in file we have. See CMake documentation at https://cmake.org/cmake/help/v3.18/command/configure_file.html?highlight=cmakedefine. Instead of the `add_definitions` and the ifdef in defines.h, I think you would only need this in config.h.in:
`#cmakedefine DEBUG_LEVEL "@LIBRARY_DEBUG_LEVEL@"`

argilo:
OK, I went the config.h.in route. Your suggested syntax didn't quite work because `DEBUG_LEVEL` is not defined in cmake. I just renamed everything to `LIBRARY_DEBUG_LEVEL`, which is probably clearer anyway.

awesie (approved):

argilo:
Oops, one small bug -- the quotes around the debug level were making it a string. Fixed directly on the master branch in https://github.com/theori-io/nrsc5/commit/26c614f5296eec95bd554ef1d4719c0da907d2f4, and the setting is now working as expected.

PR #245, opened 2020-08-26
Enhanced hdc

awesie:
Do not merge.

This is a first draft of the enhanced stream support for AM. I would like more test data but I neither have a way to record AM nor am I in the right country currently.

My testing so far has been the AM sample that was provided a while ago. The converted form that works with nrsc5 can be found here: https://gofile.io/d/lQuUUN.

For testing, there is a CLI switch (--enhanced) that controls whether the core stream will be decoded (default) or core+enhanced. This would go away before this pull request is accepted.

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -476,7 +476,7 @@ void frame_process(frame_t *st, size_t length)
     while (offset < audio_end - RS_CODEWORD_LEN)
     {
         unsigned int start = offset;
-        unsigned int j, lc_bits, loc_bytes, prog;
+        unsigned int j, lc_bits, loc_bytes, prog, seq;
         unsigned short locations[MAX_AUDIO_PACKETS];
         frame_header_t hdr = {0};
         hef_t hef = {0};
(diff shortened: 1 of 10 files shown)


src/frame.c
```diff
@@ -542,7 +543,8 @@ void frame_process(frame_t *st, size_t length)
             }
             else
             {
-                input_pdu_push(st->input, st->buffer + offset, cnt, prog, hdr.stream_id);
+                input_pdu_push(st->input, st->buffer + offset, cnt, prog, hdr.stream_id, seq);
+                seq = (seq + 1) & 0x3f;
```
argilo, line 547: The sequence number also needs to be incremented in the case where the CRC is invalid. This line won't be reached due to the `continue` on line 523.

awesie: Fair point. Looking over this code, I feel like there are similar problems with `st->pdu_idx[prog][hdr.stream_id]` as well, so some refactoring might be warranted.

argilo: Yes, you're right. We should reset `st->pdu_idx[prog][hdr.stream_id]` back to zero in the `j == 0 && hdr.pfirst` case, regardless of whether the CRC is valid.

Issue #239, opened 2020-08-13
Decode enhanced audio stream


argilo:
Currently the AM receiver only decodes the core audio stream. Decoding the enhanced audio stream would improve the audio quality since it would bring the bit rate from 20 kbps up to 36 kbps. We'll need to figure out how the enhanced audio packets are encoded.

pclov3r:
@argilo I'm sure the NRSC-5 spec covers this but maybe this is helpful anyhow? https://www.eetimes.com/understanding-hd-radio-the-program-audio-chain-part-2/

awesie:
I'm looking at this a bit, but it is more difficult than the core audio since having a separate enhanced stream is not part of HE-AACv2.

From what I can tell so far, HDC block types 3, 4, 5, 6, and 7 can all have "enhanced data". We have already seen 5 and 6 in the AM data. 7 should be stereo data. 3 and 4 could be used with FM enhanced audio.

This enhanced data is not "just" SBR data though. In block types 5 and 6, there may be SBR data in the enhanced stream, but there is still other channel data before that. I'll continue working on it and post any additional revelations.

argilo:
Interesting! I'm glad to see you're having a peek at the codec.

The other thing we'll need to do is align the Enhanced packets with the Core packets using the sequence numbers in the PDU headers. I'll take a crack at that soon.

awesie:
Still trying to understand how this works.

The enhanced stream is parsed similar to the core stream, and if the block type in the core stream was 5 or 6, then the enhanced stream is parsed as stereo (whereas the core stream is mono). Once both streams are parsed, you now have 3 channels worth of spectral data (audio in the frequency domain). The core stream channel is merged with the enhanced stream channels so you end with 2 channels.

Parsing the enhanced stream is not hard since it is exactly the same as the core stream. I'm just not sure about the merge step. I don't know enough about audio compression to know if there is an obvious way to combine these frequency domain channels. Scalable AAC (N.B. not SLS) has been a part of the AAC spec since 1999, but I don't think anyone really used it until DRM used ER AAC-Scalable. I'll continue investigating to see if DRM code is insightful.

awesie:
[...]

PR #246, opened 2020-09-03
Add support for HDC block type 7

argilo:
As @awesie noted in https://github.com/theori-io/nrsc5/issues/239#issuecomment-679113341, HDC block type 7 should be treated as stereo. I've made that change here, and can now decode audio from KSL.


awesie (approved):
good enough for now.

PR #247, opened 2020-09-04
Change faad2 upstream and update to the latest

argilo:
The faad2 project seems to be alive again at https://github.com/knik0/faad2, and Debian is now using this fork. So I think it's worth switching nrsc5 over as well.

It seems that instead of using `--with-drm` to add DRM support, the project now just builds a libfaad_drm library in parallel with libfaad. I've mirrored this pattern and added a libfaad_hdc library which has HDC support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,19 +217,19 @@ if (USE_FAAD2)
     ExternalProject_Add (
         faad2_external
         TIMEOUT 120
-        GIT_REPOSITORY "https://github.com/dsvensson/faad2.git"
-        GIT_TAG b7aa099fd3220b71180ed2b0bc19dc6209a1b418
+        GIT_REPOSITORY "https://github.com/knik0/faad2.git"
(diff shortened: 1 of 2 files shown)


awesie:
LGTM. Are you going to pull the latest version of faad2 that includes your cleanup PR?

argilo:
Yes indeed. Apparently I forgot to push that. I've done so now.

awesie:
I spent some more time looking at faad2. A couple thoughts:

1. faad2 is very much still unmaintained. The original authors stopped working on it. The new repository author disappeared and some other guy is doing the thankless work of trying to keep it alive since he is the only one with commit privileges.

2. Some of the code in upstream faad2 is utter shit. I was trying to fix a couple of their open security issues and after spending some time looking at sbr_qmf.c, I became lost in despair. The code base could use a thorough cleaning.

3. The most maintained AAC decoder seems to be fdk-aac, since it is used in Android. It is probably impossible to get HDC support upstream, though. Next best option is probably ffmpeg (libavcodec) built-in aac decoder.

PR #256, opened 2021-02-07
libnrsc5 API doxygen documentation

farlies:
Inline Doxygen documentation in nrsc5.h for libnrsc5 API functions and data structures.  Cmake option to build the documentation as html (OFF by default).  This represents a stab based on my own study/needs, so it is possible some details are inaccurate (but I would be delighted to make corrections!)


awesie:
Thanks for putting this together! It will take a few days before I have time to thoroughly review the docs.

argilo:
Thanks! This looks useful. I've squashed the commits and manually merged.

If there are errors in the documentation, we can fix those in future pull requests.

PR #270, opened 2022-02-02
Inline neon fix

coagentpai:
Compilation was failing on RPi 4 in Alpine container with neon enabled due to the __always_inline attribute. This change makes the compilation succeed with no warnings. 

I admit I am not an expert here so feedback is welcome of course.


coagentpai:
Hmm looks like the PR has some extra commits. Closing.

awesie:
Thanks. I'll commit a fix for this. I'm just going to remove `__always_inline` since it really isn't necessary.

Issue #260, opened 2021-03-21
Weird audio distortion on not-main channels


MB6:
Hello,
I've compiled nrsc5 on windows, on multiple pc's now, and when I try to tune not-main ie (1-3) subchannels, I get a weird distortion, grainy quality.

Here is a sample. https://mega.nz/file/jp1CQYjB#mCDIBDRvp3UqZIyKop3Zt2fRGqIIBRHRpxhpBv6inTc

64 bit build. I know my way around computers, but programming is not my forte. I'll provide anything else someone might need.

pclov3r:
I just noticed this now on a iHeart station subchannel here.

Is this a regression of https://github.com/theori-io/nrsc5/issues/212 however it appears that this bug was fixed?

@argilo Any opinion?

argilo:
It could also be a bug in faad2 or nrsc5's HDC patches. Unfortunately I know very little about the codec; @awesie might have more insight.

Roman-Port:
I'm also able to reproduce this. It sounds like it's dropping samples to me. Doesn't seem to affect HD-1, only HD-2 and beyond. I did some tests on my machine with the same IQ file. I built on both Windows and Ubuntu 20.02 to see if it was Windows related, but the output audio file matches byte-for-byte on the two platforms. 

Here's an example IQ recording. To reproduce, download this file and unzip it. Begin listening to HD-2 by invoking with command ``nrsc5 -r example_trim 1`` (NRSC5 starts subchannels at 0, their usual labels start at 1). After the ad and legal ID, you'll hear a song that has an especially garbled intro. Probably ~10 seconds into the song, press the 0 key to switch to HD-1 and compare it. The same ad will be playing on HD-1 and even that sounds clearer than the HD-2 channel. Once a song begins, it's a night and day difference.

https://drive.google.com/file/d/1DBqvxH8wEjrWa3nGDIFPF5TGWAvQRLhb/view?usp=sharing

Below, I've also uploaded the raw HE-AAC audio data for both HD-1 and HD-2. Note that these won't be playable without the patch in this repo.

[KPFX-HD1.aac](https://github.com/theori-io/nrsc5/files/7931052/KPFX-HD1.zip)
[KPFX-HD2.aac](https://github.com/theori-io/nrsc5/files/7931053/KPFX-HD2.zip)

TheDaChicken:
[...]

PR #273, opened 2022-02-28
Disable parametric stereo

awesie:
Parametric Stereo (PS) is used to simulate stereo audio on low bandwidth
channels. Our decoding seems to cause audio distortions (#260) so
temporarily disable PS until we can fix the implementation.


pclov3r:
@argilo 

I wonder if this would be better off as a command flag or will it just be left a compile option since the goal is to fix it.

argilo:
I'm fine with just disabling it for now.

I tested the fix with @Roman-Port's sample (from https://github.com/theori-io/nrsc5/issues/260#issuecomment-1020848180), and all sounds good now.

PR #276, opened 2022-04-18
Allow constant piped IQ data in the API

Albert-S-Briscoe:
Gnuradio blocks get constant data, which would need to be copied to a modifiable buffer before being passed to nrsc5. In src/input.c, the data gets copied to a buffer (or converted to complex) before anything else happens, so making it constant doesn't break anything within nrsc5. I don't see how it would affect anything already using the API.

nrsc5 (raw input and rtl_sdr input), cli.py (both sample formats), and nrsc5-gui work for me.


argilo:
Thanks for your contribution, and sorry for the log delay in reviewing it.

Changing these buffers to `const` seems like a good idea.

PR #281, opened 2022-06-28
Fix loading of SSE Viterbi decoder input

### LINKED ISSUE #280: Fedroa 36 constant  Lost synchronization
Hello,

I run Fedora 36 and noticed that stations don't play anymore and I get this output consistently. I have very strong signal and no issues on Ubuntu 20.04. Is this a bug or is there a fix? I followed the build instructions to the letter.  
```
20:07:45 MER: 15.7 dB (lower), 15.7 dB (upper)
20:07:45 BER: 0.501401, avg: 0.500109, min: 0.498725, max: 0.501401
20:07:45 Lost synchronization
20:07:45 Synchronized
```

Thank You.

argilo:
Fixes #280.

When nrsc5 is compiled with GCC 12 and `-DUSE_SSE=1`, Viterbi decoding fails. A compiler warning points to the `_val` array as being the source of the problem. I expect that this is because casting it to a `double *` results in undefined behaviour. I've updated the code to manually duplicate the 8-bit inputs and then load them using an unaligned SSE load. After this change, `support/sample.xz` can be decoded properly.

I don't know much about SSE programming, so I'd appreciate a review from an expert. I'd also appreciate some help with testing @andrewfer000 @pclov3r @markjfine.

### DIFF
--- a/src/conv_sse.h
+++ b/src/conv_sse.h
@@ -251,7 +251,7 @@ static inline void _sse_metrics_k7_n4(const int16_t *val, const int16_t *out,
 			    m8, m9, m10, m11, m12, m13, m14, m15)
 
 	/* (BMU) Load and expand 8-bit input out to 16-bits */
-	m7 = _mm_castpd_si128(_mm_loaddup_pd((double const *) val));
+	m7 = _mm_loadu_si128((__m128i *) val);
 
 	/* (BMU) Load and compute branch metrics */
 	m0 = _mm_load_si128((__m128i *) &out[0]);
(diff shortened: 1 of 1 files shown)


markjfine:
Unfortunately, I'm still on FC35 and gcc 11.3.1.

pclov3r:
This seems to compile and work without issues on Debian 11 which uses GCC 10.2.1 and cross compiling for windows using Mingw-w64.

markjfine:
Hadn't thought about retro-builds. I'll try rebuilding on my platforms (gcc 11.3.1 Linux, clang 12.0.0 Mac/mingw cross-build Win) later today.

markjfine:
Rebuilt for Mac and Win11:
1. No issues with Mac at all.
2. MingW cross-build works on my Parallels Win11 VM with some hiccups, but I'm pretty sure the problems are all Parallels. They borked their VM engine, which goes catatonic intermittently, as well as the USB interface, when they added ARM support - it's a disaster. nrsc5.exe will synch and you will get audio, but the audio periodically unsynchs, then stops and restarts again. Pretty sure if this were the original problem it would unsynch immediately after synching and not generate any audio at all.

Will try FC35 later today for completeness, but I'm pretty sure it's good.

markjfine:
Just to confirm, FC35 using gcc 11.3.1 is ok.

pclov3r:
Hum, Is this safe to merge at this point @argilo ? :) 

Seems to be mostly problem free at least form the testing done on various compilers.

awesie (approved):

PR #285, opened 2022-09-11
Fix distortion with parametric stereo and re-enable parametric stereo

logiconcepts819:
The proposed commit fixes the audio distortion problem with parametric stereo streams reported by #260 and also re-enables parametric stereo for HDC.  It also updates faad2 to the latest version.

Feel free to let me know if you see any additional changes in the DRM parametric stereo decoder that could be needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,7 +220,7 @@ if (USE_FAAD2)
         faad2_external
         TIMEOUT 120
         GIT_REPOSITORY "https://github.com/knik0/faad2.git"
-        GIT_TAG f71b5e81f563d94fa284977a326520d269d8353e
+        GIT_TAG df42c6fc018552519d140e3d8ffe7046ed48b0cf
         PREFIX ${FAAD2_PREFIX}
 
         UPDATE_COMMAND ""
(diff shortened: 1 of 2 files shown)


TheDaChicken:
Yep, I definitely can say that this fixed the issue!

pclov3r:
@argilo 

This has been here awhile. Is this good enough to merge? 

Would like to see this fixed :)

argilo:
I was hoping @awesie might comment, since he knows a lot more about the codec than I do. But since it's been a while, I'll have a peek at this soon and get it merged.

awesie:
Looking at the code quickly, it seems reasonable enough. If it fixes the audio distortions, then I would say go for it. I would have to do a lot more digging to check its accuracy.

TheDaChicken:
Well, we have no clue if it fixed the issue for other people as it did with me... I assume it did.

Ah, I wish HDRadio didn't sound crap on the subchannels....

argilo:
I had a look over the faad2 changes, and they seems reasonable to me. I also tested with the I/Q recording from https://github.com/theori-io/nrsc5/issues/260#issuecomment-1020848180, and the audio sounds reasonable. Merging.

Issue #287, opened 2022-12-08
Five channels?


markjfine:
Unsure if this is a mistake, but WWWT in North-Central Virginia is now showing a total of 5 HD channels:
![Screenshot 2022-12-08 at 8 55 59 AM](https://user-images.githubusercontent.com/28349323/206464519-9b91de11-45a7-432d-8afc-ebc2dd5c8cbc.jpg)

HD4 isn't generating any audio in nrsc5, so who knows what's really going on. Of course, trying to access HD5 in nrsc5 is a non-starter.

markjfine:
will do this afternoon.

markjfine:
@argilo Too big for Github, so I left it at https://fineware-swl.com/downloads/wwwt_221208180200.tar.gz (217Mb).

argilo:
I get "Access Denied" when I open that URL.

markjfine:
yeah... hang on. My security system.

markjfine:
try https://fineware-swl.com/downloads/wwwt_221208180200.iq (218Mb). As if gzipping helped shrink it much...

argilo:
Got it, thanks.

markjfine:
👍🏼

argilo:
Interesting! The station is using Primary Service Mode 11. I've only ever seen modes 1 and 3 used in the field, and those are the only modes currently supported by nrsc5. We'll have to add support for mode 11 to get HD4 and HD5 decoding.

argilo:
In case you're interested, the differences between the various service modes are described here:

https://www.nrscstandards.org/standards-and-guidelines/documents/standards/nrsc-5-d/reference-docs/1011s.pdf

markjfine:
Didn't even think to look at the service mode, just thought maybe someone at the station set something wrong and mistakenly put HD4 into some fictional HD5.

markjfine:
Also means I'll have to add a new tab/info column on dui.

argilo:
Because service mode 11 is backwards compatible with service mode 3, the following change is enough to get HD4 working:
```diff
diff --git a/src/sync.c b/src/sync.c
index e00d567..a4a9488 100644
--- a/src/sync.c
+++ b/src/sync.c
@@ -464,7 +464,7 @@ void sync_process_fm(sync_t *st)
                     decode_push_pm(&st->input->decode, DEMOD(cimagf(c)) * mult_ub);
                 }
             }
-            if (st->psmi == 3) {
[...]

PR #288, opened 2022-12-09
Add primary service mode backward compatibility

argilo:
As noted in #287, nrsc5 fails to decode the P3 logical channel for stations that are using primary service mode 11 (MP11). This happens because primary service mode backward compatibility ([Layer 1 FM](https://www.nrscstandards.org/standards-and-guidelines/documents/standards/nrsc-5-d/reference-docs/1011s.pdf), section 6.2.1) is not implemented.

I've implemented that here, and the P3 logical channel should now be decoded for modes 3 and 11 (as well as reserved modes 27, 35, 43, 51, and 59; if any of these are added in the future, they will include a backwards compatible P3 logical channel).

After this change, the HD4 channel works in @markjfine's recording (https://github.com/theori-io/nrsc5/issues/287#issuecomment-1343161505).

### DIFF
--- a/src/sync.c
+++ b/src/sync.c
@@ -29,6 +29,14 @@
 #define PARTITION_WIDTH 19
 #define MIDDLE_REF_SC 30 // midpoint of Table 11-3 in 1011s.pdf
 
+// Table 6-4 in 1011s.pdf
+static const int compatibility_mode[64] = {
+    0, 1, 2, 3, 1, 5, 6, 5, 6, 1, 2, 11, 1, 5, 6, 5,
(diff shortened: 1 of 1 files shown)


argilo:
Oops, it looks like CI is broken (for unrelated reasons -- `dlopen` isn't looking in `/usr/local/lib/` to find `libnrsc5.dylib`). I'll see about getting that fixed.

awesie (approved):
Diff looks reasonable to me. Only nitpick would be that the global variable `compatibility_mode` could be const and static.

argilo:
Thanks for the review! I'll rebase once CI is fixed, and throw in those improvements.

PR #294, opened 2023-01-29
Fix Homebrew instructions

MikeRich88:
You need the .rb or it does not work (for me anyway, on MacPro5,1 running Monterey).

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ Follow the Ubuntu instructions above, but replace the first command with the fol
 
 ## Building on macOS using [Homebrew](https://brew.sh)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,8 +34,8 @@ Follow the Ubuntu instructions above, but replace the first command with the fol
 
 ## Building on macOS using [Homebrew](https://brew.sh)
 
-    $ curl https://raw.githubusercontent.com/theori-io/nrsc5/master/nrsc5.rb > /tmp/nrsc5
-    $ brew install --HEAD -s /tmp/nrsc5
+    $ curl -O https://raw.githubusercontent.com/theori-io/nrsc5/master/nrsc5.rb > /tmp/nrsc5.rb
```
argilo, line 37: I suspect the `-O` here is unnecessary.

Issue #279, opened 2022-06-20
Long WAV files can cut off


don-code:
I recently used the `-o wav-output` feature to capture a five-hour segment of audio. The resulting WAV file cuts off at three hours, 22 minutes. If I import as raw PCM, the full five hours of audio is heard, so all samples have been dumped to the file - the size section just has not been updated. I stopped the capture with Ctrl+C - maybe there is a more appropriate way to stop?

I have not tried another five-hour (or more than 3:22) capture, but could easily do so if that'd be helpful.

Here's the last few lines of output before I stopped the capture:
```
22:58:38 MER: 12.7 dB (lower), 13.6 dB (upper)
22:58:38 BER: 0.000290, avg: 0.000663, min: 0.000016, max: 0.009602
22:58:38 Title: Family Friendly Weekends Start at 6am on Saturday
22:58:38 Artist: Discover 88.9 and WERS.ORG
22:58:38 Audio bit rate: 59.1 kbps
```

Here's the output of `ffprobe` on the output file:
```
Input #0, wav, from 'out.wav':
  Duration: 03:22:53.94, bitrate: 2064 kb/s
  Stream #0:0: Audio: pcm_s16le ([1][0][0][0] / 0x0001), 44100 Hz, stereo, s16, 1411 kb/s
```

Here are the first 100 bytes of the file, showing the header (the full file is 3.1gb):
```
00000000  52 49 46 46 f7 ff ff 7f  57 41 56 45 66 6d 74 20  |RIFF....WAVEfmt |
00000010  28 00 00 00 fe ff 02 00  44 ac 00 00 10 b1 02 00  |(.......D.......|
00000020  04 00 10 00 16 00 10 00  03 00 00 00 01 00 00 00  |................|
00000030  00 00 10 00 80 00 00 aa  00 38 9b 71 64 61 74 61  |.........8.qdata|
00000040  bb ff ff 7f 21 f4 f7 e5  ca f7 7c ec 2f 01 3b f3  |....!.....|./.;.|
00000050  85 0a 5f fb 6c 0d 39 05  9f 08 27 0f 9e 00 c5 16  |.._.l.9...'.....|
00000060  51 fb 58 1b                                       |Q.X.|
```

Here's what I get if I import as "raw PCM" in Audacity with the settings set to 16-bit stereo PCM, just to prove all of the data is indeed dumped and listenable:
[...]

PR #296, opened 2023-04-05
Adding ability to output raw pcm format in addition to wav

trs79:
Due to Issue #279 regarding wav output stopping after about 3 hours  I added a flag "-P" to allow raw pcm output which should bypass any wav file limitations. It does require that you manually set the format and sample rate for consuming programs (with ffmpeg you can use s16le for format). I'm using this command to pipe to icecast2 with great results so far:

`nrsc5 -P -q -o - 94.3 0 | ffmpeg -re -ar 44100 -vn -ac 2 -channel_layout stereo -f s16le -i - -f ogg -c:a flac  -content_type 'application/ogg' icecast://source:PASSWORD@127.0.0.1:8000/radio`

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,9 @@ Replace `32` with `64` if you want a 64-bit build. Once the build is complete, c
                                       (example: localhost:1234)
     -r iq-input                     read IQ samples from input file
     -w iq-output                    write IQ samples to output file
-    -o audio-output                 write audio to output WAV file
+    -o audio-output                 write audio to output file
+    -t audio-type                   type of audio output (wav or raw)
+                                      (default is wav. used in conjunction with -o)
     -q                              disable log output
     -l log-level                    set log level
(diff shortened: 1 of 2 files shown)


argilo:
This seems like a good idea. Since libao supports raw output, we may as well use it.

I could believe there might be future interest in adding other output formats, so perhaps it would be best to use a command-line option like `-t file-type`, where `file-type` would (for now) be either `wav` or `raw`, with `wav` being the default.

trs79:
> This seems like a good idea. Since libao supports raw output, we may as well use it.
> 
> I could believe there might be future interest in adding other output formats, so perhaps it would be best to use a command-line option like `-t file-type`, where `file-type` would (for now) be either `wav` or `raw`, with `wav` being the default.

Good idea, I changed to using -t file-type. Please let me know if there is anything else.

argilo:
Thanks for making that change.

I did a bit of cleanup and squashed the commits. I'll merge once CI is happy.

PR #301, opened 2023-06-29
Log the primary service mode after synchronization

argilo:
For debugging purposes, it's useful to know what the current primary service mode is, since some stations are beginning to use modes other than 1 and 3. For instance, in #287 we found that WWWT was using mode 11, and I recently noticed that WAMU is using mode 2 (which is not yet supported by nrsc5).

Here I've added a debug log message so that we'll notice new modes more quickly.

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -287,7 +287,10 @@ void input_set_sync_state(input_t *st, unsigned int new_state)
     if (st->sync_state == SYNC_STATE_FINE)
         nrsc5_report_lost_sync(st->radio);
     if (new_state == SYNC_STATE_FINE)
+    {
         nrsc5_report_sync(st->radio);
+        log_debug("Primary service mode: %d", st->sync.psmi);
+    }
 
     st->sync_state = new_state;
 }


markjfine:
Let me know if you need help with WAMU captures as I get it locally.

argilo:
Thanks. The recording I have should be fine for now, but I'd be interested to hear whether it's still using mode 2 today. If you pass the `-DLIBRARY_DEBUG_LEVEL=1` option to cmake, you'll be able to see the new log message.

markjfine:
Sure is:
```
14:44:45 Synchronized
Primary service mode: 2
14:44:45 Station name: WAMU

```

argilo:
Thanks for confirming. At some point I'll take a crack at decoding the P3 logical channel. There aren't too many differences between modes 2 & 3, so it probably won't be a lot of work.

In my recording of WAMU, there are only two audio programs and they're both in the main (P1) logical channel. So I'm curious what the P3 channel is used for.

markjfine:
If we weren't curious we'd never be in this hobby...

Unfortunately I don't get a good enough signal in the car to trigger HDRadio on WAMU, just in case there was something that radio could see. Next time I'm closer to Northern VA I'll try to remember to look.

PR #302, opened 2023-06-29
Implement checksum for Station Message

argilo:
The "Station Message" payload has a checksum, but nrsc5 was ignoring it up until now. Here I've added checksum validation. I verified that it works correctly with all stations in my collection of 85 recordings.


argilo:
> I verified that it works correctly with all stations in my collection of 85 recordings.

I take that back. I just noticed that one of the recordings has an invalid checksum. (It's off by one.) I'll investigate (and update this PR if necessary) before merging.

argilo:
Fixed.

The station in question had a special character in the message with the high bit set. I just needed to adjust the checksum calculation to operate on `unsigned char` values instead of (signed) `char`.

PR #306, opened 2023-07-04
Add support for primary service mode MP2

### LINKED ISSUE #305: Add support for primary service mode MP2
As noted in #301, WAMU is using primary service mode MP2, but nrsc5 cannot decode this mode yet.

I've added preliminary support in the [add-mp2 branch](https://github.com/theori-io/nrsc5/compare/add-mp2), and will open a pull request once I've had a chance to do some more testing.

It turns out that WAMU is using the P3 logical channel to send its station logos:

1003_SLWAMU$010001Logo
5002_SLWAMUHD020001Logo

My recording doesn't have any album art, but it's possible the P3 logical channel might also be used for that.

/cc @markjfine

argilo:
Fixes #305.

This pull request does the following:

1. Adds support for primary service mode MP2. This mode is very similar to mode MP3, except that the P3 logical channel is only half as wide. Because of this difference, it also uses slightly different interleaver parameters. To accomodate this mode, I've added length arguments to some of the functions that process the P3 logical channel.
2. Adds support for fixed data in logical channels other than P1. To accommodate this, we need one CCC data structure per logical channel, and various functions have to pass around the current logical channel number.
3. Adds support for L2 PDUs containing only fixed data (and no audio).

All three enhancements were needed to properly receive WAMU, which uses mode MP2 and puts its station logos in data-only L2 PDUs in the P3 logical channel.

### DIFF
--- a/src/conv.h
+++ b/src/conv.h
@@ -30,7 +30,7 @@ struct lte_conv_code {
 
 int nrsc5_conv_decode_p1(const int8_t *in, uint8_t *out);
 int nrsc5_conv_decode_pids(const int8_t *in, uint8_t *out);
-int nrsc5_conv_decode_p3_p4(const int8_t *in, uint8_t *out);
+int nrsc5_conv_decode_p3_p4(const int8_t *in, uint8_t *out, int len);
 int nrsc5_conv_decode_e1(const int8_t *in, uint8_t *out, int len);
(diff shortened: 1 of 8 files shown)


argilo:
The update code appears to work well with all my recordings, and valgrind did not identify any memory corruption bugs. Merging.

PR #321, opened 2023-07-09
Put the new expiry_utc member at the end of the lot struct

argilo:
In #310 a new member was added to the middle of the `lot` struct, which inadvertently broke [nrsc5-gui](https://github.com/cmnybo/nrsc5-gui), which has a copy of the `nrsc5.py` Python API shim. By moving the `expiry_utc` member to the end of the struct, binary compatibility can be preserved.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -328,9 +328,9 @@ struct nrsc5_event_t
             unsigned int lot;
             unsigned int size;
             uint32_t mime;
-            struct tm *expiry_utc;
             const char *name;
             const uint8_t *data;
+            struct tm *expiry_utc;
         } lot;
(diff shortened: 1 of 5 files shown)


markjfine:
Found one more fix on line 478 of nrsc5.py:
            evt = LOT(lot.port, lot.lot, MIMEType(lot.mime), expiry_time, self._decode(lot.name), lot.data[:lot.size])
should probably be:
            evt = LOT(lot.port, lot.lot, MIMEType(lot.mime), self._decode(lot.name), lot.data[:lot.size], expiry_time)
likewise moving the expiry to the end.

markjfine:
np... just noticed that this morning.

PR #325, opened 2023-10-07
Quick fix

TheDaChicken:
Any `nrsc5_open_*` function runs the worker thread.  

Because of that, using ``nrsc5_open_pipe`` would create a stale worker_thread.
Using a custom method of reading SDR and then using `nrsc5_pipe_samples_cs16`, the worker_thread loops, wastes CPU cycles & increases CPU usage.

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -195,7 +195,7 @@ static void *worker_thread(void *arg)
     return NULL;
 }
 
-static void nrsc5_init(nrsc5_t *st)
+static void nrsc5_init(nrsc5_t *st, int run_worker)
 {
     st->closed = 0;
     st->stopped = 1;
@@ -208,6 +208,9 @@ static void nrsc5_init(nrsc5_t *st)
 
     output_init(&st->output, st);
     input_init(&st->input, st, &st->output);
+
+    if(!run_worker)
+    	return;
 
     // Create worker thread
     pthread_mutex_init(&st->worker_mutex, NULL);
@@ -308,7 +311,7 @@ NRSC5_API int nrsc5_open(nrsc5_t **result, int device_index)
     err = rtlsdr_set_offset_tuning(st->dev, 1);
     if (err && err != -2) goto error;
 
-    nrsc5_init(st);
+    nrsc5_init(st, 1);
 
     *result = st;
     return 0;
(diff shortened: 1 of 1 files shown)


argilo:
> the worker_thread loops, wastes CPU cycles & increases CPU usage

Does it? When I tested this, I just saw the worker thread waiting here, and not using any CPU:

https://github.com/theori-io/nrsc5/blob/6cbc10649aa3c780407ad18c3560d9ef142f280b/src/nrsc5.c#L151-L152

Could you show what you did to cause CPU usage? I'd like to understand what's happening before making changes.

argilo:
Oh, did you call `nrsc5_start` (which starts the worker thread)? That's not necessary for the pipe mode, and I do see that it results in the worker thread going into a busy loop when using pipe mode.

argilo:
It turns out the `cli.py` example makes this exact mistake when the `-f` option is used, causing one CPU core to be pegged. It should probably skip calling `start` and `stop`.

The `nrsc5_start` and `nrsc5_stop` functions in `nrsc5.c` should probably be updated to do nothing in the pipe case.

Skipping thread allocation in `nrsc5_init` as you've done here may still have some value, since it saves allocating an unused thread object. However, in addition to `nrsc5_start` and `nrsc5_stop` we'd also need to update `nrsc5_set_callback` and `nrsc5_close` to not touch the uninitialized thread resources.

argilo:
I've opened #327 as an alternative. Let me know if it fixes your issue.

TheDaChicken:
> Does it?

I found this issue by using perf. I had unusual CPU usage on my program and didn't understand where it came from. 

> Oh, did you call `nrsc5_start` (which starts the worker thread)?

Oh crap! I do call `nrsc5_start`. 

There are also bugs I want to mention that I also found:
- `fftwf_cleanup` is called in input.c which causes any plans to be deleted. It is helpful when using Valgrind. But it kills created plans outside of the library when calling the API. [here's this](https://github.com/FFTW/fftw3/issues/176#issuecomment-522286722)
- This could be caused by weird gain issues. I am unsure if this is worth looking into? Audio frames can make crackling noises sometimes. listening to FM & while nrsc5 is trying to gather a signal, I don't think there is a good way of testing gain. I would be changing gain of FM too? 
- Losing the signal causes audio frames to first not appear. I don't know when & should audio frames come in. There is nothing that tells me that there is an underrun and when I should switch back to FM?

argilo:
Closing in favour of #327.

argilo:
> There are also bugs I want to mention that I also found:

Please open separate issues for those.

PR #327, opened 2023-10-07
Don't touch the worker thread in pipe mode

argilo:
This is an alternative to #325.

As noted there, if `nrsc5_start` is accidentally called after entering pipe mode with `nrsc5_open_pipe`, then the (unnecessary) worker thread enters into a busy loop. Here I've updated the API functions to skip all worker thread code in the pipe mode, so that API users can safely call `nrsc5_start` regardless of the mode. It will simply do nothing in the pipe mode.

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -106,6 +106,11 @@ error:
     return ret;
 }
 
+static int using_worker(nrsc5_t *st)
+{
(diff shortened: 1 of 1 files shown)


TheDaChicken:
Fixes the problem for me! 👍better fix than my quick-fix.

Issue #322, opened 2023-07-10
Coexistence between nrsc5 and other rtl-sdr tuning apps


markjfine:
This is meant as a long-term enhancement suggestion, which may or may not already be planned.

nrsc5 in itself is an amazing application for stand-alone demodulation of HD Radio signals. There are also good user interfaces available to make it useful to the average radio user. Not a knock against nrsc5, but a couple of things that's really missing are:
1.  A good way to find/detect stations that contain the HD Radio signature front and back porches within your local signal environment; and
2. A good way to find the best antenna orientation and maximize the signal strength. Right now you can only use the individual (or combined) BER values and manually tweak the gain within nrsc5 (which is cumbersome) as an estimate once it's already been run and locked onto an HD signal.

One possible way is to use something like GQRX to first tune around and then re-orient an antenna for maximum signal and reduced adjacent interference. Then run nrsc5 to decode the HD Radio portion.

The problem is ownership of the rtl-sdr: Would be nice to leave GQRX running, disengage DSP processing, then switch to something like nrsc5-gui, and hit play. The problem is that GQRX already 'owns' the USB port to the rtl-sdr and you get the message:
```
Kernel driver is active, or device is claimed by second instance of librtlsdr.
In the first case, please either detach or blacklist the kernel module
(dvb_usb_rtl28xxu), or enable automatic detaching at compile time.
```
The only way to solve it is to kill GQRX before starting nrsc5 (or nrsc5-gui/dui), thus losing the immediacy of a good visual tuning element (and hoping any multipath effects don't change things in the meantime).

Realise that adding the guts nrsc5 is out of scope for GQRX (would imagine European users then requesting a DAB+ module as well). Likewise, adding the visualization similar to that used by GQRX to nrsc5-gui/dui would likely overtax Python's ability to function.

This logically leaves the ability to have these applications be able to simultaneously run, being able to the 'see' the rtl-sdr receiver, though not necessarily be able to simultaneously operate it (e,g, nrsc5 play while GQRX DSP is engaged, and vice versa).

TheDaChicken:
I feel the same way. It would be cool to see a spectrum analyzer.

I don't think its possible for 2 applications to be accessing rtl-sdr.
I assume that adding a gui to this is out of the guts for nrsc5. 

If the API would expose iq samples then it would be possible to create a program w/ a gui using the library that has the spectrum analyzer. 
It's not fully out of my scope to use some gui library in c++ to create a gui. I ain't a gui designer.

[...]

PR #328, opened 2023-10-07
Remove input length restrictions from pipe API

argilo:
As noted in https://github.com/theori-io/nrsc5/issues/322#issuecomment-1694611094, it's annoying for API users that complex samples must be passed into `nrsc5_pipe_samples_cu8` two at a time (i.e. four bytes at a time). Here I've removed all length restrictions from `nrsc5_pipe_samples_cu8` and `nrsc5_pipe_samples_cs16` so that it's no longer even necessary to ensure that the I & Q pieces of a sample are passed in at the same time. Any leftover samples or pieces thereof are stored in a buffer until more input arrives.

@TheDaChicken Let me know if this simplifies things on your end.

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -548,14 +548,57 @@ NRSC5_API void nrsc5_set_callback(nrsc5_t *st, nrsc5_callback_t callback, void *
 
 NRSC5_API int nrsc5_pipe_samples_cu8(nrsc5_t *st, const uint8_t *samples, unsigned int length)
 {
-    input_push_cu8(&st->input, samples, length);
+    unsigned int sample_groups;
+
+    while (st->leftover_u8_num > 0 && length > 0)
+    {
+        st->leftover_u8[st->leftover_u8_num++] = samples[0];
+        samples++;
+        length--;
+
+        if (st->leftover_u8_num == 4) {
+            input_push_cu8(&st->input, st->leftover_u8, 4);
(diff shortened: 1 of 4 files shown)


TheDaChicken:
I originally wrote this in the "Quick Fix" pull request but I think it is more suited here.
I was originally saying "Audio frames can make crackling noises sometimes. That might be something to look into"

I do want to emphasize the bugginess I am getting. Underruns / cracking noises. I am converting from complex to uint8_t. I did do a complex to uint8_t (I don't think I am doing that properly due to alignment) to prevent resampling again in nrsc5 after resampling. I thought this could be caused by weird gain issues. 

Here is an example: https://youtu.be/XMswC9g92ZY

This is something I am doing for converting from float to uint8_t: 
```c++
void IQConverterFUC::Process(const DataBuffer<std::complex<float>> &input, DataBuffer<uint8_t> &output)
{
    output.SetSize(input.GetSize() * 2);
    for (int i = 0; i < input.GetSize(); ++i)
    {
        auto I = static_cast<int64_t>((input[i].real() * 128.0f) + 127.4f);
        auto Q = static_cast<int64_t>((input[i].imag() * 128.0f) + 127.4f);
        output[2 * i] = I;
        output[2 * i + 1] = Q;
    }
}
```

argilo:
> I do want to emphasize the bugginess I am getting.

Please open a separate issue for each bug you have observed.

PR #342, opened 2024-04-01
Remove non-standard MP11 interleaving

argilo:
This was added in d96b72c304f1902b4a762b632c8f56da6675f64a to deal with WWWT's non-standard signal.

It's necessary to remove it to receive WTUE's MP11 signal, which follows the standard. I would hope that WWWT has already fixed their signal, but if not I suppose we could consider adding a command-line option to select the non-standard interleaving.

/cc @markjfine

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -268,7 +268,7 @@ void decode_process_p3_p4(decode_t *st, interleaver_iv_t *interleaver, int8_t *v
         if ((out % 6) == 1 || (out % 6) == 4) // depuncture, [1, 0, 1, 1, 0, 1]
             viterbi[out++] = 0;
 
-        interleaver->internal[interleaver->i] = interleaver->buffer[interleaver->buffer_number][i];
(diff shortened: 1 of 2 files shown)


markjfine:
I never got around to making a recording yesterday - sorry about that.

That said, tested the changes to decode.c and decode.h against WWWT, which is still apparently interleaving. It's possible there's an issue with the firmware in whatever box they're using to encode and inject into the baseband prior to transmit.

Adding a command line switch may be a good idea. I wouldn't listen to the channel they're feeding HD5 in particular, but there's always the possibility other stations may use that box.

argilo:
> Adding a command line switch may be a good idea.

I started going down this path, but it would require adding an extra function to the API. I'm not sure it's worth doing that just to accommodate one station's broken transmitter. If more stations show up with the same bug, maybe we could reconsider.

markjfine:
No worries. I mean, you already raised this issue to Dave Kolesar and he didn't seem to feel it was an issue worth tackling, IRC. So, it is what it is.

Issue #330, opened 2023-10-08
Underruns & cracking noises


TheDaChicken:
I am getting: Underruns / cracking noises. 
Here is a video: https://youtu.be/XMswC9g92ZY
I could send anything like IQ recording etc of any type in the process.

Here is my code model:
- Getting SDR samples of complex<floats>
- Resampling to NRSC5's constant SAMPLE_RATE
- Converting to uint8_t for nrsc5.
- Passing it into `nrsc5_pipe_samples_cu8`

This is what I am doing for converting from float to uint8_t:
```c++
void IQConverterFUC::Process(const DataBuffer<std::complex<float>> &input, DataBuffer<uint8_t> &output)
{
    output.SetSize(input.GetSize() * 2);
    for (int i = 0; i < input.GetSize(); ++i)
    {
        auto I = static_cast<uint8_t>((input[i].real() * 128.0f) + 127.4f);
        auto Q = static_cast<uint8_t>((input[i].imag() * 128.0f) + 127.4f);
        output[2 * i] = I;
        output[2 * i + 1] = Q;
    }
}
```

argilo:
I would guess that this has the same root cause as #261 (and closely related #177).

argilo:
Basically, the library needs to implement proper buffering & delay, and when given input samples it should produce output samples at a fixed ratio. It will be a fair amount of work to get it all right, which is why I haven't taken a stab at it yet.

TheDaChicken:
That would be helpful. PortAudio LOVES to grab samples before they are available. Even just exposing some stuff related to buffering could help.

TheDaChicken:
Plus another thing I was thinking about: 

Should buffering be implemented in client code or be implemented in the API? The API can just expose stuff related to buffering like latency etc. That way buffering can be managed in any way needed.

argilo:
The client code could do some additional buffering of its own if it really needs to, but there will have to be some buffering on the API side to get the latency correct and allow things like analog blending.

TheDaChicken:
> The client code could do some additional buffering of its own if it really needs to, but there will have to be some buffering on the API side to get the latency correct and allow things like analog blending.

The reason i asked that question is because my program does the analog "blending" itself since it's not implemented in nrsc5. There is no way for me to tell any latency. Plus, the amount of silence I have to add for no-blending feels kind of weird.

TheDaChicken:
I want to make your life easier because you may not have much free time.

[...]

PR #343, opened 2024-04-24
Elastic Buffer

TheDaChicken:
Fixing #330 

Use a circular buffer to store packets to be played later temporarily. This allows libnrsc5 to handle buffering properly by outputting at a fixed ratio. This allows missing packets to be silenced or handled properly. 

~~Replace audio callback with a buffering system that involves a polling function instead with any buffer size + implement latency buffering.~~ 

~~New functions:~~
~~1. `nrsc5_open_program`~~
~~2. `nrsc5_close_program`~~
~~3. `nrsc5_reset_program`~~
~~4. `nrsc5_read_program_blocking`~~
~~5. `nrsc5_read_program_nonblocking`~~

Please let me know what you think. Code review as much as you want! :D

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -26,8 +26,6 @@
 #define PCI_AUDIO_FIXED_OPP 0x8D8D33
 #define PCI_FIXED 0x3634CE
 
-#define MAX_AUDIO_PACKETS 64
-
 typedef struct
 {
     unsigned int codec;
@@ -288,6 +286,32 @@ static unsigned int calc_lc_bits(frame_header_t *hdr)
     }
 }
 
+static unsigned int calc_avg_packets(frame_header_t *hdr)
+{
+    switch(hdr->codec)
+    {
+        case 0:
+            return 32;
+        case 1:
+        case 2:
+        case 3:
+            if (hdr->stream_id == 0)
+                return 4;
+            else
+                return 32;
+        case 10:
+            if (hdr->stream_id == 0)
+                return 32;
+            else
+                return 4;
+        case 13:
+            return 4;
+        default:
+            log_warn("unknown codec field (%d)", hdr->codec);
+            return 32;
+    }
+}
+
 static unsigned int parse_location(uint8_t *buf, unsigned int lc_bits, unsigned int i)
 {
     if (lc_bits == 16)
@@ -337,7 +361,7 @@ static void aas_push(frame_t *st, uint8_t* psd, unsigned int length, logical_cha
     else
     {
         // remove protocol and fcs fields
-        input_aas_push(st->input, psd + 1, length - 3);
+        output_aas_push(st->input->output, psd + 1, length - 3);
     }
 }
 
@@ -503,7 +527,7 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
     while (offset < audio_end - RS_CODEWORD_LEN)
     {
         unsigned int start = offset;
-        unsigned int j, lc_bits, loc_bytes, prog;
+        unsigned int j, lc_bits, loc_bytes, prog, avg, seq;
         unsigned short locations[MAX_AUDIO_PACKETS];
         frame_header_t hdr = {0};
         hef_t hef = {0};
@@ -535,6 +559,10 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
         if (hdr.hef)
             offset += parse_hef(st->buffer + offset, audio_end - offset, &hef);
         prog = hef.prog_num;
+        avg = calc_avg_packets(&hdr);
+        seq = (hdr.seq - hdr.pfirst) % MAX_AUDIO_PACKETS;
+
+        output_align(st->input->output, prog, hdr.stream_id, hdr.pdu_seq, hdr.latency, avg, seq, hdr.nop);
 
         parse_hdlc(st, aas_push, st->psd_buf[prog], &st->psd_idx[prog], MAX_AAS_LEN, st->buffer + offset, start + hdr.la_location + 1 - offset, lc);
         offset = start + hdr.la_location + 1;
@@ -555,7 +583,7 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
                     if (crc == 0)
                     {
                         memcpy(&st->pdu[prog][hdr.stream_id][idx], st->buffer + offset, cnt);
-                        input_pdu_push(st->input, st->pdu[prog][hdr.stream_id], cnt + idx, prog, hdr.stream_id);
+                        output_push(st->input->output, st->pdu[prog][hdr.stream_id], cnt + idx, prog, hdr.stream_id, seq);
                     }
                     st->pdu_idx[prog][hdr.stream_id] = 0;
                 }
@@ -576,11 +604,12 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
             {
                 if (crc == 0)
                 {
-                    input_pdu_push(st->input, st->buffer + offset, cnt, prog, hdr.stream_id);
+                    output_push(st->input->output, st->buffer + offset, cnt, prog, hdr.stream_id, seq);
                 }
             }
 
             offset += cnt + 1;
+            seq = (seq + 1) % MAX_AUDIO_PACKETS;
         }
     }
 
--- a/src/frame.c
+++ b/src/frame.c
@@ -562,7 +562,7 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
         avg = calc_avg_packets(&hdr);
         seq = (MAX_AUDIO_PACKETS + hdr.seq - hdr.pfirst) % MAX_AUDIO_PACKETS;
 
-        output_align(st->input->output, prog, hdr.stream_id, hdr.pdu_seq, hdr.latency, avg, seq, hdr.nop);
+        output_align(st->input->output, prog, hdr.stream_id, hdr.pdu_seq, hdr.latency, avg, seq);
 
         parse_hdlc(st, aas_push, st->psd_buf[prog], &st->psd_idx[prog], MAX_AAS_LEN, st->buffer + offset, start + hdr.la_location + 1 - offset, lc);
         offset = start + hdr.la_location + 1;
--- a/src/frame.c
+++ b/src/frame.c
@@ -560,7 +560,7 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
             offset += parse_hef(st->buffer + offset, audio_end - offset, &hef);
         prog = hef.prog_num;
         avg = calc_avg_packets(&hdr);
-        seq = (hdr.seq - hdr.pfirst) % MAX_AUDIO_PACKETS;
+        seq = (MAX_AUDIO_PACKETS + hdr.seq - hdr.pfirst) % MAX_AUDIO_PACKETS;
 
         output_align(st->input->output, prog, hdr.stream_id, hdr.pdu_seq, hdr.latency, avg, seq, hdr.nop);
 
--- a/src/main.c
+++ b/src/main.c
@@ -41,13 +41,13 @@
 #include "log.h"
 
 #define AUDIO_BUFFERS 128
-#define AUDIO_THRESHOLD 40
-#define AUDIO_DATA_LENGTH 8192
+#define MAX_AUDIO_DATA_LENGTH 31072
 
 typedef struct buffer_t {
     struct buffer_t *next;
     // The samples are signed 16-bit integers, but ao_play requires a char buffer.
-    char data[AUDIO_DATA_LENGTH];
+    char data[MAX_AUDIO_DATA_LENGTH];
+    size_t count;
 } audio_buffer_t;
 
 typedef struct {
@@ -143,8 +143,9 @@ static void push_audio_buffer(state_t *st, unsigned int program, const int16_t *
     st->free = b->next;
     pthread_mutex_unlock(&st->mutex);
 
-    assert(AUDIO_DATA_LENGTH == count * sizeof(data[0]));
+    assert(MAX_AUDIO_DATA_LENGTH >= count * sizeof(data[0]));
     memcpy(b->data, data, count * sizeof(data[0]));
+    b->count = count;
 
     pthread_mutex_lock(&st->mutex);
     if (program != st->program)
@@ -160,9 +161,6 @@ static void push_audio_buffer(state_t *st, unsigned int program, const int16_t *
     else
         st->head = b;
     st->tail = b;
-
-    if (st->audio_ready < AUDIO_THRESHOLD)
-        st->audio_ready++;
 
     pthread_cond_signal(&st->cond);
 
@@ -846,7 +844,7 @@ int main(int argc, char *argv[])
         audio_buffer_t *b;
 
         pthread_mutex_lock(&st->mutex);
-        while (!st->done && (st->head == NULL || st->audio_ready < AUDIO_THRESHOLD))
+        while (!st->done && (st->head == NULL))
             pthread_cond_wait(&st->cond, &st->mutex);
 
         // exit once done and no more audio buffers
@@ -863,7 +861,7 @@ int main(int argc, char *argv[])
             st->tail = NULL;
         pthread_mutex_unlock(&st->mutex);
 
-        ao_play(st->dev, b->data, sizeof(b->data));
+        ao_play(st->dev, b->data, b->count * sizeof(int16_t));
 
         pthread_mutex_lock(&st->mutex);
         // add to free list
--- a/src/output.c
+++ b/src/output.c
@@ -60,10 +60,90 @@ static unsigned int elastic_writable(elastic_buffer_t *elastic)
         return elastic->size - (elastic->write - elastic->read) - 1;
 }
 
+#ifdef USE_FAAD2
+static unsigned int decoder_buffer_write_available(decoder_t *st)
+{
+    if (st->read > st->write)
+        return (st->read - st->write) - 1;
+    else
+        return OUTPUT_BUFFER_LENGTH - (st->write - st->read) - 1;
+}
+
+static unsigned int decoder_buffer_read_available(decoder_t *st)
+{
+    if (st->write >= st->read)
+        return st->write - st->read;
+    else
+        return OUTPUT_BUFFER_LENGTH - (st->read - st->write);
+}
+
+static void decoder_buffer_write(decoder_t *dec, int16_t *buffer, unsigned int samples)
+{
+    if (decoder_buffer_write_available(dec) < samples)
+    {
+        log_error("Internal writer clock bug. Full of samples");
+        return;
+    }
+
+    if (dec->write + samples > OUTPUT_BUFFER_LENGTH)
+    {
+        unsigned int len = OUTPUT_BUFFER_LENGTH - dec->write;
+        memcpy(dec->output_buffer + dec->write, buffer, len * sizeof(*buffer));
+        memcpy(dec->output_buffer, buffer + len, (samples - len) * sizeof(*buffer));
+        dec->write = samples - len;
+    }
+    else
+    {
+        memcpy(dec->output_buffer + dec->write, buffer, samples * sizeof(*buffer));
+        dec->write = (dec->write + samples) % OUTPUT_BUFFER_LENGTH;
+    }
+}
+
+static void decoder_buffer_read(decoder_t *dec, int16_t *buffer, unsigned int samples)
+{
+    if (dec->read + samples > OUTPUT_BUFFER_LENGTH)
+    {
+        unsigned int first = OUTPUT_BUFFER_LENGTH - dec->read;
+        memcpy(buffer, &dec->output_buffer[dec->read], first * sizeof(*buffer));
+        memcpy(&buffer[first], dec->output_buffer, (samples - first) * sizeof(*buffer));
+        dec->read = samples - first;
+    }
+    else
+    {
+        memcpy(buffer, &dec->output_buffer[dec->read], samples * sizeof(*buffer));
+        dec->read = (dec->read + samples) % OUTPUT_BUFFER_LENGTH;
+    }
+}
+
+static unsigned int output_align_source(decoder_t *dec, unsigned int input, int all_round)
+{
+    unsigned int iq_upper, frames;
+
+    iq_upper = (input * 8);
+    if(!all_round)
+        iq_upper += dec->leftover;
+    frames = (iq_upper / 135);
+
+    if (all_round)
+    {
+        if (iq_upper % 135 > 0)
+            frames += 1;
+    }
+    else
+    {
+        dec->leftover = (iq_upper % 135);
+    }
+
+    return frames * AUDIO_FRAME_CHANNELS;
+}
+#endif
+
 static void elastic_decode_packet(output_t *st, unsigned int program, int16_t **audio, unsigned int *frames)
 {
-    decoder_t *dec = &st->decoder[program];
-    elastic_buffer_t *elastic = &dec->elastic_buffer;
+    elastic_buffer_t *elastic = &st->elastic[program];
+#ifdef USE_FAAD2
+    decoder_t *dec = &elastic->decoder;
+#endif
 
     if (elastic->ptr[elastic->read].size != 0)
     {
@@ -116,8 +196,7 @@ static void elastic_decode_packet(output_t *st, unsigned int program, int16_t **
 
 void output_align(output_t *st, unsigned int program, unsigned int stream_id, unsigned int pdu_seq, unsigned int latency, unsigned int avg, unsigned int seq, unsigned int nop)
 {
-    decoder_t *dec = &st->decoder[program];
-    elastic_buffer_t *elastic = &dec->elastic_buffer;
+    elastic_buffer_t *elastic = &st->elastic[program];
     unsigned int forward;
 
     if (stream_id != 0)
@@ -152,6 +231,9 @@ void output_align(output_t *st, unsigned int program, unsigned int stream_id, un
         log_debug("Elastic buffer created. Program: %d, Size %d bytes, Read %d pos, Write: %d pos", program, elastic->size, elastic->read, elastic->write);
     }
 
+#ifdef USE_FAAD2
+    decoder_t *dec = &elastic->decoder;
+
     if (!dec->output_buffer)
     {
         dec->output_buffer = malloc(OUTPUT_BUFFER_LENGTH * sizeof(*dec->output_buffer));
@@ -163,6 +245,7 @@ void output_align(output_t *st, unsigned int program, unsigned int stream_id, un
         else
             dec->write = (FFTCP_AM * 256 / 135) * AUDIO_FRAME_CHANNELS;
     }
+#endif
 
     // Re-sync (lost-synchronization with reader and writer)
     forward = compute_forward_sequence_position(elastic, seq);
@@ -173,74 +256,19 @@ void output_align(output_t *st, unsigned int program, unsigned int stream_id, un
     }
 }
 
-static unsigned int output_buffer_writeable(decoder_t *st)
-{
-    if (st->read > st->write)
-        return (st->read - st->write) - 1;
-    else
-        return OUTPUT_BUFFER_LENGTH - (st->write - st->read) - 1;
-}
-
-unsigned int output_buffer_available(decoder_t *st)
-{
-    if (st->write >= st->read)
-        return st->write - st->read;
-    else
-        return OUTPUT_BUFFER_LENGTH - (st->read - st->write);
-}
-
-static void output_buffer_write(decoder_t *dec, int16_t *buffer, unsigned int samples)
-{
-    if (output_buffer_writeable(dec) < samples)
-    {
-        log_error("Internal writer clock bug. Full of samples");
-        return;
-    }
-
-    if (dec->write + samples > OUTPUT_BUFFER_LENGTH)
-    {
-        unsigned int len = OUTPUT_BUFFER_LENGTH - dec->write;
-        memcpy(dec->output_buffer + dec->write, buffer, len * sizeof(*buffer));
-        memcpy(dec->output_buffer, buffer + len, (samples - len) * sizeof(*buffer));
-        dec->write = samples - len;
-    }
-    else
-    {
-        memcpy(dec->output_buffer + dec->write, buffer, samples * sizeof(*buffer));
-        dec->write = (dec->write + samples) % OUTPUT_BUFFER_LENGTH;
-    }
-}
-
-static void output_read_buffer(decoder_t *dec, int16_t *buffer, unsigned int samples)
-{
-    if (dec->read + samples > OUTPUT_BUFFER_LENGTH)
-    {
-        unsigned int first = OUTPUT_BUFFER_LENGTH - dec->read;
-        memcpy(buffer, &dec->output_buffer[dec->read], first * sizeof(*buffer));
-        memcpy(&buffer[first], dec->output_buffer, (samples - first) * sizeof(*buffer));
-        dec->read = samples - first;
-    }
-    else
-    {
-        memcpy(buffer, &dec->output_buffer[dec->read], samples * sizeof(*buffer));
-        dec->read = (dec->read + samples) % OUTPUT_BUFFER_LENGTH;
-    }
-}
-
 void output_advance_elastic(output_t *st, int pos, unsigned int used)
 {
     for (int i = 0; i < MAX_PROGRAMS; i++)
     {
-        decoder_t *dec = &st->decoder[i];
-        elastic_buffer_t *elastic = &dec->elastic_buffer;
+        elastic_buffer_t *elastic = &st->elastic[i];
         unsigned int sample_avg = average_acquire_samples(st, elastic);
 
         // Skip if no buffer
         if (!elastic->ptr)
             continue;
 
-        if (elastic->pos == -1)
-            elastic->pos = pos;
+        if (elastic->iq_pos == -1)
+            elastic->iq_pos = pos;
 
         // Packet clock
         elastic->clock += (int)used;
@@ -255,7 +283,7 @@ void output_advance_elastic(output_t *st, int pos, unsigned int used)
             {
                 elastic_decode_packet(st, i, &audio, &decoded_frames);
 #ifdef USE_FAAD2
-                output_buffer_write(dec, audio, decoded_frames);
+                decoder_buffer_write(&elastic->decoder, audio, decoded_frames);
 #endif
             }
             elastic->clock -= (int)sample_avg;
@@ -264,34 +292,14 @@ void output_advance_elastic(output_t *st, int pos, unsigned int used)
     }
 }
 
-static unsigned int output_align_source(decoder_t *dec, unsigned int input, int all_round)
-{
-    unsigned int iq_upper, frames;
-
-    iq_upper = (input * 8);
-    if(!all_round)
-        iq_upper += dec->leftover;
-    frames = (iq_upper / 135);
-
-    if (all_round)
-    {
-        if (iq_upper % 135 > 0)
-            frames += 1;
-    }
-    else
-    {
-        dec->leftover = (iq_upper % 135);
-    }
-
-    return frames * AUDIO_FRAME_CHANNELS;
-}
-
 void output_advance(output_t *st, unsigned int len)
 {
+#ifdef USE_FAAD2
+    // Output audio samples
     for (int i = 0; i < MAX_PROGRAMS; i++)
     {
-        decoder_t *dec = &st->decoder[i];
-        elastic_buffer_t *elastic = &dec->elastic_buffer;
+        elastic_buffer_t *elastic = &st->elastic[i];
+        decoder_t *dec = &elastic->decoder;
         int16_t *audio_frame;
         unsigned int hd_samples, delay_samples;
         unsigned int audio_frames, silence_frames, frames_len;
@@ -302,9 +310,9 @@ void output_advance(output_t *st, unsigned int len)
 
         // Program started in the middle of the sample.
         // Insert silence to makeup for it. It takes time to generate samples
-        if (elastic->pos > 0)
+        if (elastic->iq_pos > 0)
         {
-            hd_samples = (len - elastic->pos);
+            hd_samples = (len - elastic->iq_pos);
             delay_samples = (len - hd_samples);
         }
         else
@@ -320,26 +328,26 @@ void output_advance(output_t *st, unsigned int len)
         audio_frame = malloc(frames_len * sizeof(*audio_frame));
         memset(audio_frame, 0, silence_frames * sizeof(*audio_frame));
 
-        if (output_buffer_available(dec) < audio_frames)
+        if (decoder_buffer_read_available(dec) < audio_frames)
         {
-            log_error("Missing output samples. Requested: %d Available: %d", audio_frames, output_buffer_available(dec));
-            audio_frames = output_buffer_available(dec);
+            log_error("Missing output samples. Requested: %d Available: %d", audio_frames, decoder_buffer_read_available(dec));
+            audio_frames = decoder_buffer_read_available(dec);
         }
 
-        output_read_buffer(dec, audio_frame + silence_frames, audio_frames);
+        decoder_buffer_read(dec, audio_frame + silence_frames, audio_frames);
         nrsc5_report_audio(st->radio, i, audio_frame, frames_len);
 
         free(audio_frame);
 
         // Reset
-        elastic->pos = -1;
+        elastic->iq_pos = -1;
     }
+#endif
 }
 
 void output_push(output_t *st, uint8_t *pkt, unsigned int len, unsigned int program, unsigned int stream_id, unsigned int seq)
 {
-    decoder_t *dec = &st->decoder[program];
-    elastic_buffer_t *elastic = &dec->elastic_buffer;
+    elastic_buffer_t *elastic = &st->elastic[program];
 
     if (stream_id != 0)
         return; // TODO: Process enhanced stream
@@ -399,24 +407,15 @@ void output_reset(output_t *st)
 {
     aas_reset(st);
 
-#ifdef USE_FAAD2
     for (int i = 0; i < MAX_PROGRAMS; i++)
     {
-        decoder_t *dec = &st->decoder[i];
-        elastic_buffer_t *buffer = &dec->elastic_buffer;
+        elastic_buffer_t *buffer = &st->elastic[i];
 
+#ifdef USE_FAAD2
+        decoder_t *dec = &buffer->decoder;
         if (dec->aacdec)
             NeAACDecClose(dec->aacdec);
         dec->aacdec = NULL;
-
-        buffer->write = 0;
-        buffer->read = 0;
-        buffer->clock = 0;
-        buffer->pos = -1;
-
-        if (buffer->ptr)
-            free(buffer->ptr);
-        buffer->ptr = NULL;
 
         dec->leftover = 0;
         dec->write = 0;
@@ -426,8 +425,17 @@ void output_reset(output_t *st)
         if (dec->output_buffer)
             free(dec->output_buffer);
         dec->output_buffer = NULL;
-    }
 #endif
+
+        buffer->write = 0;
+        buffer->read = 0;
+        buffer->clock = 0;
+        buffer->iq_pos = -1;
+
+        if (buffer->ptr)
+            free(buffer->ptr);
+        buffer->ptr = NULL;
+    }
 }
 
 void output_init(output_t *st, nrsc5_t *radio)
@@ -437,8 +445,8 @@ void output_init(output_t *st, nrsc5_t *radio)
 #ifdef USE_FAAD2
     for (int i = 0; i < MAX_PROGRAMS; i++)
     {
-        st->decoder[i].aacdec = NULL;
-        st->decoder[i].leftover = 0;
+        st->elastic[i].decoder.aacdec = NULL;
+        st->elastic[i].decoder.leftover = 0;
     }
 
     memset(st->silence, 0, sizeof(st->silence));
--- a/src/output.c
+++ b/src/output.c
@@ -86,10 +86,12 @@ static int decoder_buffer_write(decoder_t *dec, const int16_t *buffer, unsigned 
 
     if (dec->write + samples > OUTPUT_BUFFER_LENGTH)
     {
-        unsigned int len = OUTPUT_BUFFER_LENGTH - dec->write;
-        memcpy(dec->output_buffer + dec->write, buffer, len * sizeof(*buffer));
-        memcpy(dec->output_buffer, buffer + len, (samples - len) * sizeof(*buffer));
-        dec->write = samples - len;
+        const unsigned int available = OUTPUT_BUFFER_LENGTH - dec->write;
+        const unsigned int remaining = samples - available;
+
+        memcpy(dec->output_buffer + dec->write, buffer, available * sizeof(*buffer));
+        memcpy(dec->output_buffer, buffer + available, remaining * sizeof(*buffer));
+        dec->write = remaining;
     }
     else
     {
@@ -107,10 +109,12 @@ static int decoder_buffer_read(decoder_t *dec, int16_t *buffer, unsigned int sam
 
     if (dec->read + samples > OUTPUT_BUFFER_LENGTH)
     {
-        unsigned int first = OUTPUT_BUFFER_LENGTH - dec->read;
-        memcpy(buffer, &dec->output_buffer[dec->read], first * sizeof(*buffer));
-        memcpy(&buffer[first], dec->output_buffer, (samples - first) * sizeof(*buffer));
-        dec->read = samples - first;
+        const unsigned int available = OUTPUT_BUFFER_LENGTH - dec->read;
+        const unsigned int remaining = samples - available;
+
+        memcpy(buffer, &dec->output_buffer[dec->read], available * sizeof(*buffer));
+        memcpy(&buffer[available], dec->output_buffer, remaining * sizeof(*buffer));
+        dec->read = remaining;
     }
     else
     {
@@ -133,12 +137,12 @@ static unsigned int output_align_source(decoder_t *dec, unsigned int input, int 
     if(!round)
         iq_num += dec->leftover;
 
-    frames = iq_num / 135;
+    frames = iq_num / RADIO_NUM;
 
-    if (round && iq_num % 135 > 0)
+    if (round && iq_num % RADIO_NUM > 0)
         frames += 1;
-    else if (round)
-        dec->leftover = iq_num % 135;
+    else if (!round)
+        dec->leftover = iq_num % RADIO_NUM;
 
     return frames * AUDIO_FRAME_CHANNELS;
 }
@@ -234,7 +238,7 @@ void output_align(output_t *st, unsigned int program, unsigned int stream_id, un
         // Align Writer (buffer->delay + seq) & Reader
         elastic_realign_forward(elastic, seq, pdu_seq, avg, seq);
 
-        log_debug("Elastic buffer created. Program: %d, Size %d bytes, Read %d pos, Write: %d pos", program, elastic->size, elastic->read, elastic->write);
+        log_debug("Elastic buffer created. Program: %d, Size %d bytes, Read %d pos, Write: %d pos, Delay: %d, Average: %d", program, elastic->size, elastic->read, elastic->write, elastic->delay, elastic->avg);
     }
     else
     {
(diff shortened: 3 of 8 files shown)


FoxxMD:
If you need testers I'm happy to build this into my workflow and see how it performs

argilo:
Thanks for the proposed fix. It looks like there is a problem with the Python wrapper, as it now crashes.

I'm not sure that such a drastic change is needed. An alternative approach would be to use a 64-frame circular buffer (as the NRSC-5 specification hints at), with silence inserted for any missing audio frames. When a batch of samples arrives through the rtlsdr callback, a corresponding batch of audio samples could immediately be returned from the circular buffer, and then decoding of the RF samples (which can take a variable amount of time) could proceed. I think this approach would result in a constant output rate, and allow the digital output to be properly synchronized with analog, without needing any change to the API.

TheDaChicken:
> Thanks for the proposed fix. It looks like there is a problem with the Python wrapper, as it now crashes.
> 

That would make sense since I didn't test the Python wrapper. Probably due to me removing the AUDIO callback. 

> I'm not sure that such a drastic change is needed. An alternative approach would be to use a 64-frame circular buffer (as the NRSC-5 specification hints at), with silence inserted for any missing audio frames. When a batch of samples arrives through the rtlsdr callback, a corresponding batch of audio samples could immediately be returned from the circular buffer, and then decoding of the RF samples (which can take a variable amount of time) could proceed. I think this approach would result in a constant output rate, and allow the digital output to be properly synchronized with analog, without needing any change to the API.

I agree about the drastic change being pretty annoying. I tried to not add silence. I'll do more testing and update the PR with changes. I know it is possible to constantly send very SMALL amounts of audio packets. However, it was VERY on edge but that was without silence. You have more experience reading whitepapers. I read it so many times.

argilo:
> You have more experience reading whitepapers. I read it so many times.

The specification only describes how to build a transmitter, so one has to infer how to build a receiver.

The "PDU control word" is what hints at a 64-frame circular buffer, with the "Starting Sequence Number" field describing where to begin inserting frames into the buffer when decoding a PDU.

The "Audio Transport" section also describes how elastic buffering is supposed to work at a high level. The receiver is supposed to use an elastic buffer so that it outputs exactly 32 audio frames per PDU, even though the incoming PDUs can have anywhere from 24 to 40 audio frames each. This is the buffer that nrsc5 is currently missing, which causes our output audio rate to be variable.

TheDaChicken:
> 
> The specification only describes how to build a transmitter, so one has to infer how to build a receiver.
> 
> The "PDU control word" is what hints at a 64-frame circular buffer, with the "Starting Sequence Number" field describing where to begin inserting frames into the buffer when decoding a PDU.

Yes. It sounds like a regular array with 64 sequence numbers. Should I assume packets are meant to be ordered per stream? It would be nice to support the alignment of many streams. I don't have a recording of any enhanced streams to test alignment.

Should I start playback once it receives all parts of the PDU Sequence and max sequence? I think delaying more would cause the audio to be more delayed than a normal receiver. 

> The "Audio Transport" section also describes how elastic buffering is supposed to work at a high level. The receiver is supposed to use an elastic buffer so that it outputs exactly 32 audio frames per PDU, even though the incoming PDUs can have anywhere from 24 to 40 audio frames each. This is the buffer that nrsc5 is currently missing, which causes our output audio rate to be variable.

If I know correctly, PDU timing is also variable. I've had a PDU take 1.8 seconds before. Since you are willing for me to touch other areas for this, it looks like `input_push` would be great to push audio frames through when it's the best time to do so.

argilo:
> Yes. Should I assume packets are meant to be ordered per stream? If its not reset buffer? It would be nice to support the alignment of many streams.

The API streams out all audio programs at once, so we'd need a separate buffer for each one.

> I don't have a recording of any enhanced streams to test alignment.

I've only seen enhanced streams used on AM. But nrsc5 doesn't currently decode them, so I think it's fine to ignore them for now. (Related: #245) If you need a recording, I can send you one.

> Should I start playback once it receives full PDU Sequence?

A real receiver would output analog audio (for the HD1 channel, anyway). nrsc5 doesn't currently demodulate the analog, but for now it could output silence, to maintain a constant input / output ratio. The same will need to be done for audio frames that are rejected due to an invalid CRC. Slots in the circular buffer could simply default to silence.

> I am not sure about 1-3 codec with 4 packets maximum.

Those are used on logical channels (e.g. P3) that have a "transfer frame modulus" of 8, i.e. 8 transfer frames per block. With an average of 4 audio frames per transfer frame, it still comes out to an average of 32 audio frames per block. The PDU control word may work a bit differently though; it's been a while since I looked at that.

> it looks like input_push would be great time to push audio packets through output.c when it's the best time to do so.

Yeah, as long as it happens before the digital demodulation steps (which can take a variable amount of time) we should be good.

argilo:
> That way, I don't need to create another thread like I was testing.

Right, I do believe it should be possible to do this without adding a thread.

TheDaChicken:
> > Should I start playback once it receives full PDU Sequence?
> 
> A real receiver would output analog audio (for the HD1 channel, anyway). nrsc5 doesn't currently demodulate the analog, but for now it could output silence, to maintain a constant input / output ratio. The same will need to be done for audio frames that are rejected due to an invalid CRC. Slots in the circular buffer could simply default to silence.
> 

Apologies. I mean, how long should I delay the HD packets before they get inserted? Currently, the program delays it by 40 because of 4 latency * 2 + 32 average packets since... the huge maximum amount of time it can take to receive enough data. I guess that's right? I want to make sure that the whole thing is alright with you if there is anything else I am wrong about.

I mean, I can see why you don't want to do this. This is painful. I don't think any of us can be sure honestly. I compared my dad's car (because mine is too old to listen to hdradio) to nrsc5. nrsc5 is ahead by ~0.3 seconds.  

 > > I don't have a recording of any enhanced streams to test alignment.
> 
> I've only seen enhanced streams used on AM. But nrsc5 doesn't currently decode them, so I think it's fine to ignore them for now. (Related: #245) If you need a recording, I can send you one.
> 

The reason I wanted a recording is because I am not sure how much delayed the enhanced stream compared to the core stream. This is because if I spend the time coding a delay. I can align it at that point. it is apart of the elastic buffer anyway.

TheDaChicken:
@argilo (Pinging for redundancy) 

Hey! I implemented a version of a circular buffer for outputting at a fixed ratio. Buffer is at the bleeding edge. 

I store packets instead of audio frames
- Sequence numbers don't equal audio frames because each packet represents _N+1_  frames due to the AAC decoder.  
- Enhanced streams can be aligned easily (since #245 uses a hack)

Basically, my buffer looks like this:
||latency|| + ||64|| + ||latency||

This allows the reader to be delayed by the latency of extra silence. I couldn't think of a way to split this into 2 different buffers. Let me know if you would rather this be split into 2 different buffers. Also, you mentioned, there needs to be 64 packets. 

The last problem with this method is the availability of the packets. They are only properly available after 32 packet time. Any audio driver needs audio before the audio frame is finished. Callback outputs at constant packet time = constant rate which causes lags at the start of playback. I will work on fixing that unless you have an idea yourself.  

Please let me know what you think

ferrellsl:
I compiled this branch under MSYS2 to do a little testing.  The BER climbs to very high rates after several minutes of play and loses sync for several seconds causing large gaps in playback.

TheDaChicken:
> The BER climbs to very high rates after several minutes of playback and loses sync for several seconds causing large gaps in playback even on very strong stations where there's no interference or weak signal.

This is either: bad gain / low signal / bad performance 

I ran a profiler just in case this new code degrades performance. The new code is very low. `acquire_process` takes up most CPU which is normal. 

If audio is dropping out with a good BER then its a new code issue and would be great to see a recording.

ferrellsl:
> > The BER climbs to very high rates after several minutes of playback and loses sync for several seconds causing large gaps in playback even on very strong stations where there's no interference or weak signal.
> 
> This is either: bad gain / low signal / bad performance
> 
> I ran a profiler just in case this new degrades performance. The new code is very low. `acquire_process` takes up most CPU which is normal.
> 
> If audio is dropping out with a good BER then its a new code issue and would be great to see a recording.

The bad gain / low signal /bad performance just doesn't make sense because I'm running an  i9-11980H CPU at 3.6Ghz on stations with a receive signal strength of 30-50db.  And yes, I used -DUSE_SSE=ON at build time.  The average BER stays in the normal range of around .02-.04% for about 10 minutes and then climbs to anywhere from 10% to 40% with the binary compiled from your branch.  The audio then drops out for 5 seconds or so and then the error rate drops back into the normal range for another minute or two with the audio track resuming for several minutes before the cycle continues a few minutes later.  

This problem doesn't not occur at all when running the binary compiled from the original theori-oi/nrsc5 repo. 

Here's a screen shot immediately after the audio drops out showing the bit rates and error rate info as well as signal strength using the binary compiled from your branch.

image

This problem occurs whether I use the binary from the command line or in conjunction with a python gui.  I've verified the same problem on another PC (i7 CPU) using a completely different SDR dongle thinking it might be a system specific problem but the problem persists across both systems.

TheDaChicken:
> This problem occurs whether I use the binary from the command line or in conjunction with a python gui. I've verified the same problem on another PC (i7 CPU) using a completely different SDR dongle thinking it might be a system specific problem but the problem persists across both systems.

I think you are seeing this issue here except its worse:
> The last problem with this method is the availability of the packets. They are only properly available after 32 packet time. Any audio driver needs audio before the audio frame is finished. Callback outputs at constant packet time = constant rate which causes lags at the start of playback. I will work on fixing that unless you have an idea yourself.

I will see what I can do.

ferrellsl:
> > This problem occurs whether I use the binary from the command line or in conjunction with a python gui. I've verified the same problem on another PC (i7 CPU) using a completely different SDR dongle thinking it might be a system specific problem but the problem persists across both systems.
> 
> I think you are seeing this issue here except its worse:
> 
> > The last problem with this method is the availability of the packets. They are only properly available after 32 packet time. Any audio driver needs audio before the audio frame is finished. Callback outputs at constant packet time = constant rate which causes lags at the start of playback. I will work on fixing that unless you have an idea yourself.
> 
> I will see what I can do.

I way out of my depth when it comes to this type of coding.  I'm creating a recording at the moment if you think it may hold some clues.

ferrellsl:
Here's a 7 minute sample.  The first audio drop occurs at 3:12 and the second at 6:25 so it appears to be happening at a very regular interval.  I can make a longer recording to confirm this if you'd like.  https://www.dropbox.com/scl/fi/u9w7bozp02wpigmtq8ob8/sample.wav?rlkey=zmk5y47vlgn7uaxq4zbvn1wl4&st=1f9bg6s0&dl=0

image

ferrellsl:
Here's a longer, completely different sample (15 minutes).  Curiously, it also experiences the first audio drop at 3:12 and again at 6:25, so there does appear to be a pattern to this.  https://www.dropbox.com/scl/fi/ofia7o5fz4m0grm568a35/sample2.wav?rlkey=c063rqif26rl0aj6ej4tz3g2p&st=ue9t9hor&dl=0

image

TheDaChicken:
> Here's a 7 minute sample. The first audio drop occurs at 3:12 and the second at 6:25 so it appears to be happening at a very regular interval. I can make a longer recording to confirm this if you'd like. https://www.dropbox.com/scl/fi/u9w7bozp02wpigmtq8ob8/sample.wav?rlkey=zmk5y47vlgn7uaxq4zbvn1wl4&st=1f9bg6s0&dl=0

Yep. I figured out the problem - I should have tested subchannels with 4 packet maximum.

Thanks for testing. Try the new latest commit.

ferrellsl:
I think you solved the problem.  There was still a drop at 30 seconds and again at 60 seconds, but after that, everything was just fine and I kept recording for 10 minutes without any further issues.

image

ferrellsl:
I ran the binary on my other test system which has an i7-4710Q CPU and the audio drop outs are back.  They seem to occur every 3 min and 12 seconds apart.  This isn't happening on my system equipped with an i9-11980HK CPU. Strange.

image

https://www.dropbox.com/scl/fi/iat6xymv8iwl2qcpx8vjh/sample5.wav?rlkey=qwx6fwxsekdgnuuxbzgixd1ql&st=31y1oyf8&dl=0
https://www.dropbox.com/scl/fi/uwc7c61f1iszixwhpv1yi/sample6.wav?rlkey=q1d86nwh71zv331eq2fxt8qk6&st=mt746qfl&dl=0

TheDaChicken:
> I ran the binary on my other test system which has an i7-4710Q CPU and the audio drop outs are back. They seem to occur every 3 min and 12 seconds apart. This isn't happening on my system equipped with an i9-11980HK CPU. Strange.

Could you send me the output of the IQ samples using -w?

ferrellsl:
> > I ran the binary on my other test system which has an i7-4710Q CPU and the audio drop outs are back. They seem to occur every 3 min and 12 seconds apart. This isn't happening on my system equipped with an i9-11980HK CPU. Strange.
> 
> Could you send me the output of the IQ samples using -w?

Here's an IQ file that's 10 minutes in length.  Everything really starts go get out of sync badly around the 7 minute mark.  https://www.dropbox.com/scl/fi/53p06s6avcq2v88ml44q0/sample.zip?rlkey=j5otk9jv625vx0o03bms1d0aw&st=j0bzgpo0&dl=0

TheDaChicken:
> > > I ran the binary on my other test system which has an i7-4710Q CPU and the audio drop outs are back. They seem to occur every 3 min and 12 seconds apart. This isn't happening on my system equipped with an i9-11980HK CPU. Strange.

It should be fixed now. The reader clock was going way too fast. Latency was incorrect. The value wasn't multiplied for the delay. 

Ignore the continued dumb commits. I deselect things on Clion. It pushes them away. I should use the terminal instead.

ferrellsl:
Thanks for the update.  This appears to have fixed things.   I still get the occasional drop out from time to time but I'm not sure if it's code related or if it's RF interference.  Thanks for all the effort you're putting into this.  Here's a screenshot of the the occasional audio drops:
image

And the IQ and WAV files:

https://drive.google.com/file/d/1pR4pjmE8DZ5Ko2L5GnTIume4zF8lVWIq/view?usp=sharing
https://drive.google.com/file/d/13sNWonAVr-Ghm_FH8CvmSTrJk1mTuv1a/view?usp=sharing

TheDaChicken:
> Thanks for the update. This appears to have fixed things. I still get the occasional drop out from time to time but I'm not sure if it's code related or if it's RF interference but I don't experience any drops or interference using the theori-io code. Thanks for all the effort you're putting into this.

It was both RF interference and code-related. I made a silly mistake. Anyway should be fixed.

ferrellsl:
There's still a code related issue.  Here's the output from the latest build.  There were only 4 audio drops during this 15 min run,  but 3 of them were at regular intervals of appox. 1 min 45 secs apart.  Getting much better though.
image

https://drive.google.com/file/d/1_uewc4lwYNZxejcp2KI4wSAHt11dLcBZ/view?usp=sharing
https://drive.google.com/file/d/1SsVFS9RC1ae8rcvLBDc7rcqa9r5gnD90/view?usp=sharing

TheDaChicken:
> There's still a code related issue. Here's the output from the latest build. There were only 4 audio drops during this 15 min run, but 3 of them were at regular intervals of appox. 1 min 45 secs apart. Getting much better though.

Thanks for the testing yet again even how repetitive this is. I am glad someone else is testing this out. The HDRadio stations here aren't super different. 

Unless I couldn't tell from the recording, it looks like RF interference or unrelated to the buffer. From the IQ file, I can see it lost synchronization and resyncs. The buffer properly resyncs tied to that while also keeping playback. 

nrsc5 still needs more things to improve reception.

ferrellsl:
> > There's still a code related issue. Here's the output from the latest build. There were only 4 audio drops during this 15 min run, but 3 of them were at regular intervals of appox. 1 min 45 secs apart. Getting much better though.
> 
> Thanks for the testing yet again even how repetitive this is. I am glad someone else is testing this out. The HDRadio stations here aren't super different.
> 
> Unless I couldn't tell from the recording, it looks like RF interference or unrelated to the buffer. From the IQ file, I can see it lost synchronization and resyncs. The buffer properly resyncs tied to that while also keeping playback.
> 
> nrsc5 still needs more things to improve reception.

You're welcome.  I've enjoyed the testing.  Looking forward to any improvements that come about as time progresses.

ferrellsl:
Just wanted to post another update after some further testing.  I recompiled after your latest commits and have let the binary run for about 12 hours now without any issues.  This is the most stable version of nrsc5 yet on my systems.  Whenever I used the binary compiled from theori-io, it would freeze or the audio would get choppy at anywhere from 15 minutes to 4 hours of playback.  I'm going to let your binary run for another 24 hours or so just to see what happens.  I don't expect any glitches though. Thanks again for the hard work you've put into your fork.  I really appreciate it.

TheDaChicken:
I made 2 changes to help the audio even better hopefully.

I may have been wrong about why lag occurred at the start (I could be wrong). It may be due to how long it takes to produce audio frames. There is a huge ratio between decoding HDRadio to waiting for 32 packet frames or 4 packet frames. I send an extra packet if there is still a little bit in the reading clock. 

This means that sometimes an extra packet is produced including when the buffer starts. This is because some of the IQ samples gets 'used' by getting put into a buffer. If the buffer is full enough then it decodes HDRadio. No matter what, it sends 'used' length to `output_advance` which increases the reading clock. 

The last 2 issues: 
- Changing the channels. There may be a race condition where you can change the channel while packets are being sent through the callback. This causes a lag spike because it doesn't start playback with all the burst packets. 
- Python cli.py has so many audio lag spikes. This may be due to a small audio buffer size. It may also be due to the overhead that exists with callbacks. ?

The fix could be for both: sending huge packets instead of many small packets.

I could be saying a bunch of nonsense. i did some stuff!

ferrellsl:
> I made 2 changes to help the audio even better hopefully.
> 
> Just to document it here:
> 
> I may have been wrong about why lag occurred at the start (I could be wrong). It may be due to how long it takes to produce audio frames. There is a huge ratio between decoding HDRadio to waiting for 32 packet frames or 4 packet frames.
> 
> To quickly combat this, I send an extra packet if there is still a little bit in the reading clock. This means that sometimes an extra packet is sent including when the buffer starts. This is because _only some_ of the IQ samples gets 'used'. Then some is stored in a buffer to be decoded. If the buffer is full enough then it decodes HDRadio. This is why previously, there was an issue with the reader clock going too fast if I used the exact IQ sample length sent to _be_ decoded. Currently, no matter what, it sends the 'used' length to `output_advance` which increases the reading clock.
> 
> I could calculate the backpressure and send more packets. That way it is fully a 'fixed-ratio' of Iq samples to audio packets.
> 
> If the buffer takes out a packet, there are fewer packets. We are stuck in the 32-packet time. Weird buffering moment.
> 
> The last 2 issues:
> 
> * Changing the channels. There may be a race condition where you can change the channel while packets are being sent through the callback. This causes a lag spike because it doesn't start playback with all the burst packets.
> * Python cli.py has so many audio lag spikes. This may be due to a small audio buffer size. It may also be due to the overhead that exists with callbacks. ?
> 
> The fix could be for both: sending huge packets instead of many small packets.
> 
> I could be saying a bunch of nonsense. i did some stuff!

Thanks for the latest commits.  They resolve an audio skipping issue when using nrsc5 with the python GUI here under MSYS2:  https://github.com/cmnybo/nrsc5-gui

Previous versions of the nrsc5 binary used with this GUI would always result in audio skipping after a variable amount of time usually ranging from 15 minutes to 4 hours after starting playback.  I had always assumed the issue was related to pyaudio under MSYS2 because the nrsc5 binary alone didn't exhibit this skipping audio behavior.  Everything has been running smooth now for about 18 hours.

The same skipping was present in the NRSC5-DUI (https://github.com/markjfine/nrsc5-dui), a more feature-rich GUI.  I'll test using NRSC5-DUI next to see if the audio skipping there has been resolved there as well and report back.

ferrellsl:
Using NRSC5-DUI, the audio skip problem still persists but I'm going to assume the problem lies with NRSC5-DUI's python code and not your code as the skipping was resolved under nrsc5-gui.  The skipping starts on my system after approximately 3 hours of play. I'm curious to know if others are having this problem with NRSC5-DUI.

markjfine:
> Using NRSC5-DUI, the audio skip problem still persists but I'm going to assume the problem lies with NRSC5-DUI's python code and not your code as the skipping was resolved under nrsc5-gui. The skipping starts on my system after approximately 3 hours of play. I'm curious to know if others are having this problem with NRSC5-DUI.

I haven't had time to test this pull, but this is odd. DUI doesn't do anything to process audio in real time and lets nrsc5 do all of it. At best, it's just parsing the status output of and interacting with nrsc5 via a command pipe. Conversely, GUI interacts and does quite a bit of audio processing in Python outside of nrsc5, in addition to processing commands/status via the API.

ferrellsl:
> > Using NRSC5-DUI, the audio skip problem still persists but I'm going to assume the problem lies with NRSC5-DUI's python code and not your code as the skipping was resolved under nrsc5-gui. The skipping starts on my system after approximately 3 hours of play. I'm curious to know if others are having this problem with NRSC5-DUI.
> 
> I haven't had time to test this pull, but this is odd. DUI doesn't do anything to process audio in real time and lets nrsc5 do all of it. At best, it's just parsing the status output of and interacting with nrsc5 via a command pipe. Conversely, GUI interacts and does quite a bit of audio processing in Python outside of nrsc5, in addition to processing commands/status via the API.

OK, the problem seems to be with rtl_tcp and/or WSL2 itself instead of your python code.  I tested this branch of NRSC5 and NRSC5_DUI on a VMware virtual machine running Ubuntu 24.04 and there were no problems at all.  

My earlier test was on a Windows 11 machine with Ubuntu running under WSL2 with my SDR dongle being hosted under 
 Windows and rtl_tcp providing the connection between the dongle and NRSC5_DUI.

markjfine:
...or something related to rtl_tcp, which makes a lot of sense to me. I had a lot of problems getting nrsc5 and dui to work reliably under Win10/11 and WSL2/MSYS and basically threw my hands up. Because everything is being emulated in that kind of environment, you're at the mercy of how well real time devices and events are handled (priority of handling interrupt calls, etc.). That said, I was using a VM of Win10/11 on a Mac at the time. When using WSL2 or MSYS, it was basically creating a *nix-like VM within a Win VM on what is already a *nix-like system to begin with (macOS being FreeBSD-based). I now have a stand-alone Lenovo Legion (mostly for gaming), so things might be a lot better. Given time I may try to take all this up again.

TheDaChicken:
> I could calculate the backpressure and send more packets. That way it is fully a 'fixed-ratio' of Iq samples to audio packets.

After lots of testing, the correct model of relying on a reader clock from decoding HDRadio itself wasn't super great at "fixed output". This made it difficult to properly send enough packets. I was relying on predicting and using calculations to determine the amount of backpressure due to decoding. 

I labeled the previous "buffer" as an "elastic buffer." The elastic buffer now writes to an output buffer. Output Buffer outputs based on IQ samples. 

This is a great:
A) I can have the audio frames to be equal to the IQ samples. 
B) This is very helpful for FM Blending in the future. 
C) Super fast channel switching (Probably even way too FAST) ??

This all in total makes aligning enhanced streams for AM, and FM blending easier. 

Anyways, I am going to go to sleep. Good night! hahaha.

TheDaChicken:
> OK, the problem seems to be with rtl_tcp and/or WSL2 itself instead of your python code

That makes more sense! I thought about it - I was so confused because the Python code for nrsc5-dui did attach the terminal pipe with buffering. So there wouldn't have been any performance issues running under Python. 

I was able to get RTLSDR running under WSL2 if I remember correctly. Though I just found running anything in a VM so much better.

markjfine:
Good stuff. Pretty sure B) was the goal. 😉

ferrellsl:
> > OK, the problem seems to be with rtl_tcp and/or WSL2 itself instead of your python code
> 
> That makes more sense! I thought about it - I was so confused because the Python code for nrsc5-dui did attach the terminal pipe with buffering. So there wouldn't have been any performance issues running under Python.
> 
> I was able to get RTLSDR running under WSL2 if I remember correctly. Though I just found running anything in a VM so much better.

RTL_TCP is sort of a dumpster fire in regard to buffering and audio skipping so I eliminated it from the equation altogether and installed USBIPD to allow nrsc5-dui to have native USB support under WSL2.  Here's the discussion that I created for nrsc5-dui using USBIPD and the instructions for enabling USB access under WSL2.  I prefer WSL2 instead of VMware or VirtualBox because they are resource hogs and WSL2 has better performance:

https://github.com/markjfine/nrsc5-dui/discussions/41

TheDaChicken:
> I labeled the previous "buffer" as an "elastic buffer." The elastic buffer now writes to an output buffer. Output Buffer can output based on IQ samples easily and has a little leeway for unique frame sizes. I'm unsure if this should be for MPS only for FM Blending.

I decided that the "output buffer" should be for MPS only/analog blending*. Not all programs need it. It should probably then be based on if blending is enabled (aka AM all digital). However, I think that's out of the scope of this PR. 

I was able to properly calculate backpressure. IQ samples equal the (rounded up) amount of audio samples outputted. If there aren't any more issues. Then to wrap this PR up it sounds like I could do something like this:
- Add something in the API to determine whether the audio is pending for a program
- If there aren't enough samples, just don't even try to insert audio to avoid: *audio* silence *audio* silence

OR the PR is finished and that can be in a separate PR. 

I am pinging @argilo to see how he feels. He hasn't given much of his opinion though he is busy. I don't blame him. It was recently the 4th of July anyways.

argilo:
> I am pinging @argilo again to see how he feels (sorry hahaha).

I've been on vacation for the past few weeks, but I should have a chance to look over this once I'm back home later this week.

TheDaChicken:
I realized that due to decimating, the ratio of AM samples to Audio frames is different than FM. This is because I am using the samples for acquiring. I fixed that. I can't test if I correctly did it. I have no way of receiving AM.

I squished previous comments to clear the huge git history.

TheDaChicken:
> I have no way of receiving AM.

Sounds like it would be useful for an AM sample test just like there is with FM. 

Anyway, I am bumping this again. Would be great to get another opinion and test. I think my code is bad. 
Please read comment: https://github.com/theori-io/nrsc5/pull/343#issuecomment-2167221070 for an explanation on this buffer.

TheDaChicken:
> > I am pinging @argilo again to see how he feels (sorry hahaha).
> 
> I've been on vacation for the past few weeks, but I should have a chance to look over this once I'm back home later this week.

Have you checked this out and have an opinion on this PR? @argilo

argilo:
Sorry for the long delay in looking at this. I've been busier than expected preparing for Defcon and the GNU Radio Conference CTF. I'll try to get this reviewed soon.

src/output.c
```diff
@@ -27,30 +27,333 @@
 #include "private.h"
 #include "unicode.h"
 
-void output_push(output_t *st, uint8_t *pkt, unsigned int len, unsigned int program, unsigned int stream_id)
+#define RADIO_FRAME_SAMPLES_FM (NRSC5_AUDIO_FRAME_SAMPLES * 135 / 8)
+#define RADIO_FRAME_SAMPLES_AM (NRSC5_AUDIO_FRAME_SAMPLES * 135 / 256)
```
argilo, line 31: AM seems to be broken; I get lots of "Internal writer clock bug. Full of samples". The intermediate sample rate for AM is 1/16 of the rate for FM, so this change seems to fix the problem:
```suggestion
#define RADIO_FRAME_SAMPLES_AM (NRSC5_AUDIO_FRAME_SAMPLES * 135 / 128)
```

TheDaChicken: Thanks for testing AM. You're correct. I made an error reducing fractions.

src/frame.c
```diff
@@ -535,6 +559,10 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
         if (hdr.hef)
             offset += parse_hef(st->buffer + offset, audio_end - offset, &hef);
         prog = hef.prog_num;
+        avg = calc_avg_packets(&hdr);
+        seq = (hdr.seq - hdr.pfirst) % MAX_AUDIO_PACKETS;
```
argilo, line 563: This calculation looks dangerous. `hdr.seq` and `hdr.pfirst` are unsigned integers. If `hdr.seq` is 0 and `hdr.pfirst` is 1, then an overflow occurs and the result of the subtraction is `UINT_MAX`. The final result is still correct since the integer modulus happens to be a multiple of `MAX_AUDIO_PACKETS`, but I think it would be safer to add `MAX_AUDIO_PACKETS` into the sum to avoid overflow.

```suggestion
        seq = (MAX_AUDIO_PACKETS + hdr.seq - hdr.pfirst) % MAX_AUDIO_PACKETS;
```

src/output.h
```diff
@@ -85,17 +90,50 @@ typedef struct
...
+    unsigned int size, read, write;
+    unsigned int latency, avg, delay;
+
+    unsigned int clock;
+    int pos;
+} elastic_buffer_t;
+
+typedef struct
+{
+    NeAACDecHandle aacdec;
```
argilo, line 113: The fact that this is defined outside `#ifdef HAVE_FAAD2` causes compilation to fail if nrsc5 is built without FAAD2 (`-DUSE_FAAD2=OFF`):

```
In file included from /home/argilo/git/nrsc5/src/input.h:13,
                 from /home/argilo/git/nrsc5/src/acquire.c:21:
/home/argilo/git/nrsc5/src/output.h:113:5: error: unknown type name ‘NeAACDecHandle’
  113 |     NeAACDecHandle aacdec;
      |     ^~~~~~~~~~~~~~
make[2]: *** [src/CMakeFiles/nrsc5_object.dir/build.make:76: src/CMakeFiles/nrsc5_object.dir/acquire.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:104: src/CMakeFiles/nrsc5_object.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

argilo: `decode_t` is only used if `HAVE_FAAD2` is defined, so I think we could safely put this whole `typedef` inside `#ifdef HAVE_FAAD2`.

src/frame.c
TheDaChicken: Yeahh... Good idea. It wouldn't hurt to add that to prevent any weird issues.

src/output.h
TheDaChicken: I swapped around. `elastic_buffer_t` contains `decoder_t` This seems to make this a lot better. I fixed the compilation.

TheDaChicken: I t4ested this and I can compile with `DUSE_FAAD2=OFF`. Am going to mark this as resolved unless you cannot reproduce this.

src/main.c
```diff
@@ -41,13 +41,13 @@
 #include "log.h"
 
 #define AUDIO_BUFFERS 128
-#define AUDIO_THRESHOLD 40
-#define AUDIO_DATA_LENGTH 8192
+#define MAX_AUDIO_DATA_LENGTH 31072
```
argilo, line 44: Where does this number come from?

src/output.h
```diff
@@ -17,6 +17,11 @@
 #define MAX_FILE_BYTES 65536
 #define MAX_LOT_FRAGMENTS (MAX_FILE_BYTES / LOT_FRAGMENT_SIZE)
 
+#define AUDIO_FRAME_CHANNELS 2
+#define AUDIO_FRAME_LENGTH (NRSC5_AUDIO_FRAME_SAMPLES * AUDIO_FRAME_CHANNELS)
+
+#define OUTPUT_BUFFER_LENGTH (64 * AUDIO_FRAME_LENGTH)
```
argilo, line 23: Where does 64 come from? Is it `MAX_AUDIO_PACKETS` or something else?

src/output.h
```diff
@@ -1,14 +1,14 @@
 #pragma once
 
 #include "config.h"
+#include "frame.h"
```
argilo, line 4: I am guessing this is here to pull in `MAX_AUDIO_PACKETS`. If so, it may probably be better to move it to `defines.h`, which has constants that are needed in more than one file.

src/main.c
TheDaChicken: Good question. This is the maximum size this PR outputs due to the SDR chuck size `(512 * 1204)` defined in nrsc5.c.
Basically: `(512 * 1024) / 1488375 * 44100 * sizeof(int)` rounded up a little bit.

src/output.h
TheDaChicken: 64 is technically an arbitrary number. I chose the output buffer to be 64 frames. This is due to the flaw that the output buffer is limited in size. I can switch to a linked list approach. That would require more code when splitting frames to output at a fixed ratio. Wanted the code to be simple.

src/output.h
TheDaChicken: Most likely a good idea + I don't think MAX_AUDIO_PACKETS constant needs to be publicly accessible outside the library.

src/output.h
argilo: OK, if it's unrelated to `MAX_AUDIO_PACKETS` then it's best to stay as it is. You could perhaps give this 64 its own name.

I'll have a look over the rest of the design to see what I think of linked list vs. buffer.

TheDaChicken:
I did some testing. Probably wasn't enough to properly test this. I wanted to manually see what it would be like aligning analog and digital.

I played back raw recordings in SDRSharp & used nrsc5 to output the audio to a file. 

I tried recordings from 4 stations. KQMV, KBKS, KHTP, KING-FM. I first tried adding `common_delay * 4` to the delay on the elastic buffer. This didn't exactly align properly with any stations except maybe KBKS. The digital audio was behind by latency + FFT delay. This could be what `after audio decoding` means on NRSC5 papers?

Not sure if this was because KBKS was the only properly aligned one out of the bunch. Other stations were either behind too much or in the future. For example, KING-FM was behind 24 frames ish on HD when adding `common_delay` Though, I read an article saying that KING-FM used a realigner box 

Assuming any of this is correct, the output buffer (buffer to temporarily store audio frames to add an extra audio frame delay + output fixed ratio based on IQ)'s size needs to be increased. Could be increased to the maximum number from `common_delay` Sounds like too big of a number though.

markjfine:
Just a thought, but as you've seen: I'm pretty sure that stations are going to have different timing between their HD streams and their FM programming. Each uses a different method of producing the stream, therefore creating variable delays - much like various TV streams have differing lags versus their on-air equivalents. I would expect that there may not be a single way to align these in all cases. You may want to just get it close and insert some 'dead air' when switching between the two. The initial buffering period may be sufficient to do this. Hopefully, it doesn't switch in and out to rapidly, which could have pretty annoying effects.

TheDaChicken:
> Just a thought, but as you've seen: I'm pretty sure that stations are going to have different timing between their HD streams and their FM programming. 

HDRadio has a parameter called `common_delay` which allows for the HD station to provide how much to delay the digital to align it properly with analog. But yeah - that doesn't mean the station producing the stream is properly handling this. I want to test the buffer whether or not it's "correct"

argilo:
Sometimes when acquiring a station, I see the following:
```
Best gain: 42.1 dB, CNR: 22.2 dB
Block @ 7
19:36:54 Synchronized
Primary service mode: 1
19:36:54 Country: US, FCC facility ID: 0
19:36:54 Station name: ABCD-FM
19:36:55 Country: US, FCC facility ID: 0
19:36:55 Station name: ABCD-FM
19:36:55 Audio program 0: public, type: News, sound experience 0
19:36:55 Country: US, FCC facility ID: 0
19:36:55 Station name: ABCD-FM
19:36:55 Station location: 40.689209, -74.044556, 96m
19:36:55 Audio program 0: public, type: News, sound experience 0
19:36:56 MER: 23.4 dB (lower), 23.4 dB (upper)
19:36:56 BER: 0.000000, avg: 0.000000, min: 0.000000, max: 0.000000
Elastic buffer created. Program: 0, Size 80 bytes, Read 32 pos, Write: 8 pos
Elastic buffer realigned. Program: 0, Read 32 pos, Write: 8 pos
19:36:56 Title: Title
19:36:56 Artist: Artist
19:36:56 XHDR: 1 BE4B7536 -1
elastic buffer full. skipped packet. bug?
elastic buffer full. skipped packet. bug?
elastic buffer full. skipped packet. bug?
elastic buffer full. skipped packet. bug?
elastic buffer full. skipped packet. bug?
elastic buffer full. skipped packet. bug?
elastic buffer full. skipped packet. bug?
elastic buffer full. skipped packet. bug?
```

argilo:
It seems to happen 50% of the time with my synthetic signal. If the buffer creation message has `Write: 8 pos` then the errors appear; when it has `Write: 40 pos` there's no error.

TheDaChicken:
Could I have a quick IQ sample? That would be helpful. Rarely I got this on stations but never enough

argilo:
Give this a try:

https://argilo.dyndns.org/files/elastic-bug.cu8

argilo:
There are a few sign-compare warnings that could be cleaned up like so:

```diff
diff --git a/src/output.c b/src/output.c
index 7c063d8..3685b4c 100644
--- a/src/output.c
+++ b/src/output.c
@@ -213,7 +213,7 @@ void output_align(output_t *st, unsigned int program, unsigned int stream_id, un
         elastic->size  = (elastic->delay * 2) + MAX_AUDIO_PACKETS;
         elastic->ptr   = malloc(elastic->size * sizeof(*elastic->ptr));
 
-        for (int i = 0; i < elastic->size; i++)
+        for (unsigned int i = 0; i < elastic->size; i++)
         {
             elastic->ptr[i].size = 0;
             elastic->ptr[i].seq = -1;
@@ -274,12 +274,12 @@ void output_advance_elastic(output_t *st, int pos, unsigned int used)
         elastic->clock += (int)used;
 
         // Decode packets based on average
-        while (elastic->clock >= (int)sample_avg)
+        while (elastic->clock >= (unsigned int)sample_avg)
         {
             int16_t *audio;
             unsigned int decoded_frames;
 
-            for (int j = 0; j < elastic->avg; j++)
+            for (unsigned int j = 0; j < elastic->avg; j++)
             {
                 elastic_decode_packet(st, i, &audio, &decoded_frames);
 #ifdef USE_FAAD2
```

TheDaChicken:
> Could I have a quick IQ sample? That would be helpful. Rarely I got this on stations but never enough

Thank you so much! I fixed it. Silly mistake. The available size was calculated using read as head. That would be normal behavior on a regular buffer. 

I couldn't get IDE autocomplete since the build doesn't work - fftw.org is down. 

> There are a few sign-compare warnings that could be cleaned up like so:

That should be fixed now. Let me know if I have made a mistake somewhere.

TheDaChicken:
In the few months I've been using it, I found having to change reader position instead of appending especially when lag occurs isn't completely a good idea. Nothing bad as occurred. Just in case, I want to end up fixing it. When I get to it, I'll implement my idea unless there are other ways. Also maybe research FM realigning a bit more to ensure the buffer is correct. 

Currently, this PR does work perfectly fine. I think some function names need to properly named and some code moved (I hate my terrible code).

argilo:
Sorry for the long delay in reviewing this. I haven't had much inspiration to work on nrsc5 lately, but I'm motivated now because I'm experimenting with integrating it into [ka9q-radio](https://github.com/ka9q/ka9q-radio).

I'm doing some testing with the `nrsc5_pipe_samples_cu8` and `nrsc5_pipe_samples_cs16` API functions, and things are mostly working as expected. But I noticed that AM mode does not work with `nrsc5_pipe_samples_cs16`. The following error message constantly appears:

```
Internal writer clock bug. Full of samples
```

TheDaChicken:
> Sorry for the long delay in reviewing this. I haven't had much inspiration to work on nrsc5 lately, but I'm motivated now because I'm experimenting with integrating it into [ka9q-radio](https://github.com/ka9q/ka9q-radio).
> 
> I'm doing some testing with the `nrsc5_pipe_samples_cu8` and `nrsc5_pipe_samples_cs16` API functions, and things are mostly working as expected. But I noticed that AM mode does not work with `nrsc5_pipe_samples_cs16`. The following error message constantly appears:
> 
> ```
> Internal writer clock bug. Full of samples
> ```

Hi, welcome back! I almost didn't see this until much later because pclover messaged me on Discord. 

Do you mind sending an AM sample? We've talked about it already but that would be very helpful to just have one. I have seen this before and I wonder if its similar issue. This is most likely a quick fix. 

The code for this isn't 100% perfect. I have been really wanting to hear from someone that is familiar with this codebase. If you have any questions, let me know.

argilo:
Try these:

https://argilo.dyndns.org/files/krld.cu8 (8-bit unsigned, 1488375 sps)
https://argilo.dyndns.org/files/krld.cs16 (16-bit signed, 46511.71875 sps)

argilo:
The Python CLI uses `nrsc5_pipe_samples_cs16`, so it's possible to reproduce the problem like so:

```
support/cli.py --am -r krld.cs16 --iq-input-format cs16 0
```

argilo:
> The 16bit file you provided isn't long enough to sync. 8-bit file had a few minutes.

They are both the same length. Maybe you didn't have the entire file. The sizes are:

* krld.cu8: 365366012 bytes
* krld.cs16: 22835240 bytes

argilo:
> Are you on an old git commit? I pushed my up-to-date code just in case.

I was on the latest. I updated again just now, and I see the new error message when testing AM + CS16.

argilo:
> Alright it should work great now!

AM CS16 works now, but AM CU8 is broken.

```
Missing output samples. Requested: 970 Available: 0
```

argilo:
Note that the CU8 sample rate is 1488375 for both FM and AM. (Chosen because it's the only power-of-two multiple of the OFDM channel spacing that's also a valid RTL-SDR sample rate.)

But for CS16, the rate is 1488375/2 for FM and 1488375/32 for AM. (Chosen because they are the smallest power-of-two multiple of the OFDM channel spacing that covers the necessary bandwidth.)

TheDaChicken:
Yep, I got it! 

I was working on trying to fix it in the best way - there isn't easily a good way. This should work with regular FM, AM cs16, AM cu8. Let me know if you have any issues!

argilo:
OK, all four combinations are working now. :+1: 

I'll put the code through its paces, then have another look over the changes, hopefully tomorrow.

TheDaChicken:
> OK, all four combinations are working now. 👍
> 
> I'll put the code through its paces, then have another look over the changes, hopefully tomorrow.

Alright, I hope you have a good night. I will make another change to make it more pretty and maybe fix a bug I noticed recently. Then leave it for you tomorrow.

TheDaChicken:
> I will make one more change to make it more pretty and maybe fix a bug

I was unable to fix this bug - it seems not like a quick fix. 

I tried to reproduce it with other subchannels in my area with 4 packet averages but couldn't seemly do it. For some reason, playing back CKFM HD4 causes an overrun on this buffer. The output buffer somehow gets overrun with so many valid audio packets?

```
02:00:31 Stream data: port=0881 seq=7F5F mime=4EB03469 size=50
02:00:31 Stream data: port=0881 seq=7F60 mime=4EB03469 size=342
02:00:31 Title: Electric
02:00:31 Artist: Shawn Desman
02:00:31 Genre:
Decoder buffer full for program: 3. Skipped samples. bug?
Decoder buffer full for program: 3. Skipped samples. bug?
Decoder buffer full for program: 3. Skipped samples. bug?
Decoder buffer full for program: 3. Skipped samples. bug?
02:00:32 Country: CA, FCC facility ID: 352
02:00:32 Station name: CKFM
02:00:32 Slogan: 99.9 Virgin Radio
02:00:32 Message: 99.9 Virgin Radio Torontos 1 Hit Music Station
02:00:32 Station location: 43.642456, -79.387207, 3488m
02:00:32 Audio program 0: public, type: Top 40, sound experience 0
02:00:32 Audio program 1: public, type: Information, sound experience 0
02:00:32 Audio program 2: public, type: Sports, sound experience 0
02:00:32 Audio program 3: public, type: Country, sound experience 0
02:00:32 Data service: public, type: Navigation, MIME type 469
02:00:32 Data service: public, type: Image Maps, MIME type 2d7
02:00:32 Data service: public, type: Image Maps, MIME type e96
02:00:32 Data service: public, type: Navigation, MIME type eb2
Decoder buffer full for program: 3. Skipped samples. bug?
02:00:32 Audio bit rate: 46.6 kbps
02:00:32 MER: 9.8 dB (lower), 9.0 dB (upper)
02:00:33 BER: 0.000233, avg: 0.000442, min: 0.000137, max: 0.004121
02:00:33 Stream data: port=0880 seq=ED19 mime=B39EBEB2 size=951
02:00:33 Stream data: port=0880 seq=ED1A mime=B39EBEB2 size=159
02:00:33 Stream data: port=0881 seq=7F61 mime=4EB03469 size=86
02:00:33 LOT file: port=0810 lot=147 name=TMT_035dso_3_2_20250101_2334_0093.png size=2128 mime=4F328CA0 expiry=2025-01-02T00:06:00Z
File TMT_035dso_3_3_20250101_2334_0092.png, size 4224, lot 146, port 0810, mime 4F328CA0
02:00:33 Title: Electric
02:00:33 Artist: Shawn Desman
02:00:33 Genre:
02:00:33 Title: Electric
02:00:33 Artist: Shawn Desman
02:00:33 Genre:
02:00:33 Country: CA, FCC facility ID: 352
02:00:33 Station name: CKFM
02:00:33 Slogan: 99.9 Virgin Radio
02:00:33 Message: 99.9 Virgin Radio Torontos 1 Hit Music Station
02:00:33 Station location: 43.642456, -79.387207, 3360m
02:00:33 Audio program 0: public, type: Top 40, sound experience 0
02:00:33 Audio program 1: public, type: Information, sound experience 0
02:00:33 Audio program 2: public, type: Sports, sound experience 0
02:00:33 Audio program 3: public, type: Country, sound experience 0
02:00:33 Data service: public, type: Navigation, MIME type 469
02:00:33 Data service: public, type: Image Maps, MIME type 2d7
02:00:33 Data service: public, type: Image Maps, MIME type e96
02:00:33 Data service: public, type: Navigation, MIME type eb2
02:00:34 Audio bit rate: 46.5 kbps
02:00:34 MER: 9.8 dB (lower), 9.0 dB (upper)
02:00:34 BER: 0.000265, avg: 0.000441, min: 0.000137, max: 0.004121
02:00:34 Stream data: port=0880 seq=ED1B mime=B39EBEB2 size=1036
02:00:34 Stream data: port=0880 seq=ED1C mime=B39EBEB2 size=476
02:00:34 Stream data: port=0881 seq=7F62 mime=4EB03469 size=248
02:00:34 Title: Electric
02:00:34 Artist: Shawn Desman
02:00:34 Genre:
02:00:34 Title: Electric
02:00:34 Artist: Shawn Desman
02:00:34 Genre:
02:00:35 Country: CA, FCC facility ID: 352
02:00:35 Station name: CKFM
02:00:35 Slogan: 99.9 Virgin Radio
02:00:35 Message: 99.9 Virgin Radio Torontos 1 Hit Music Station
02:00:35 Station location: 43.642456, -79.387207, 3472m
02:00:35 Audio program 0: public, type: Top 40, sound experience 0
02:00:35 Audio program 1: public, type: Information, sound experience 0
02:00:35 Audio program 2: public, type: Sports, sound experience 0
02:00:35 Audio program 3: public, type: Country, sound experience 0
02:00:35 Data service: public, type: Navigation, MIME type 469
02:00:35 Data service: public, type: Image Maps, MIME type 2d7
02:00:35 Data service: public, type: Image Maps, MIME type e96
02:00:35 Data service: public, type: Navigation, MIME type eb2
02:00:35 Audio bit rate: 46.6 kbps
Decoder buffer full for program: 3. Skipped samples. bug?
Decoder buffer full for program: 3. Skipped samples. bug?
Decoder buffer full for program: 3. Skipped samples. bug?
02:00:35 MER: 9.8 dB (lower), 9.0 dB (upper)
02:00:35 BER: 0.000202, avg: 0.000441, min: 0.000137, max: 0.004121
02:00:35 Stream data: port=0880 seq=ED1D mime=B39EBEB2 size=559
02:00:36 Stream data: port=0880 seq=ED1E mime=B39EBEB2 size=551
02:00:36 Stream data: port=0881 seq=7F63 mime=4EB03469 size=196
02:00:36 Title: Electric
02:00:36 Artist: Shawn Desman
02:00:36 Genre:
02:00:36 Title: Electric
02:00:36 Artist: Shawn Desman
02:00:36 Genre:
02:00:36 Country: CA, FCC facility ID: 352
02:00:36 Station name: CKFM
02:00:36 Slogan: 99.9 Virgin Radio
02:00:36 Message: 99.9 Virgin Radio Torontos 1 Hit Music Station
02:00:36 Station location: 43.642456, -79.387207, 48m
02:00:36 Audio program 0: public, type: Top 40, sound experience 0
02:00:36 Audio program 1: public, type: Information, sound experience 0
02:00:36 Audio program 2: public, type: Sports, sound experience 0
02:00:36 Audio program 3: public, type: Country, sound experience 0
02:00:36 Data service: public, type: Navigation, MIME type 469
02:00:36 Data service: public, type: Image Maps, MIME type 2d7
02:00:36 Data service: public, type: Image Maps, MIME type e96
02:00:36 Data service: public, type: Navigation, MIME type eb2
Decoder buffer full for program: 3. Skipped samples. bug?
02:00:37 Audio bit rate: 45.9 kbps
```

Link:
```
https://www.dropbox.com/scl/fi/obwuw7ta8n7m84qnywsif/CKFM-FM-1_1_25.raw?rlkey=xg5elvq6soh35t8crmwrf4xjt&st=bvrqilvk&dl=1
```

argilo:
It appears that when an HDC packet is missed (e.g. due to an invalid CRC) it results in 4096 samples of silence. I would have expected 2048.

argilo:
It appears the extra silence is related to resetting the AAC decoder. If I comment out the code that resets it, then the silent periods are 2048 samples long, as expected. Maybe the decoder has a startup delay? I don't remember the details of how it works.

argilo:
> Maybe the decoder has a startup delay?

Yeah, that seems to be the case. Resetting the decoder when there's a gap seems like a good idea. :+1:

argilo:
> For some reason, playing back CKFM HD4 causes an overrun on this buffer.

I see the same error on some of my own recordings.

In addition, several of my AM recordings produce occasional errors of this sort:

```
Missing output samples. Requested: 972 Available: 894
Missing output samples. Requested: 970 Available: 850
Missing output samples. Requested: 970 Available: 886
Missing output samples. Requested: 970 Available: 894
Missing output samples. Requested: 972 Available: 850
```

TheDaChicken:
> > For some reason, playing back CKFM HD4 causes an overrun on this buffer.
> 
> I see the same error on some of my own recordings, including one AM recording.
> 
> In addition, several of my AM recordings produce occasional errors of this sort:
> 
> ```
> Missing output samples. Requested: 972 Available: 894
> Missing output samples. Requested: 970 Available: 850
> Missing output samples. Requested: 970 Available: 886
> Missing output samples. Requested: 970 Available: 894
> Missing output samples. Requested: 972 Available: 850
> ```

Were the recordings that occurred this problem with 4 packet averages? I know that AM will be 4 packet average no matter what because of the 'core' stream. 

Do you know if you could verify some of my terrible arithmetic or make it more clear? 

For example for this line:
```
        unsigned int pushed = input_push_to_acquire(st);
        acquire_process(&st->acq);
        output_advance_elastic(st->output, (int)st->used - (int)pushed - (int)pos, pushed);
```
`pushed` should be equal to amount of samples that were used (including skipped samples?).
   - When I lose signal (unplug my antenna & plug it back in), The "elastic clock" should still be correct and allow putting audio samples in the correction location. This sadly doesn't happen because maybe wrong calculation with `st->skip`. You may see `Elastic buffer realigned. Program: 0, Read 0 pos, Write: 22 pos` which tires to get get around it. I can improve this 'resync behavior' for other issues if i am clear that the clock is working fine.  

`(int)st->used - (int)pushed - (int)pos` is meant to calculate the current position we are working with on the latest added IQ sample (another word is new IQ sample). Audio can start decoding in a middle of a huge IQ sample. This is used to add silence to prevent underruns. This is a problem with huge buffer sizes. This can be negative because we may be working with the previous IQ sample. 

Does this make sense?

argilo:
> Were the recordings that occurred this problem with 4 packet averages?

Yes, they were.

> Do you know if you could verify some of my terrible arithmetic or make it more clear?

I will have a look at your code soon. To start, I had to re-familiarize myself with the behaviour of things like `pdu_seq` and `seq` in the headers. Now that it's fresh in my mind, the code might make some sense. :smile:

TheDaChicken:
Good idea 👍

argilo:
One difference I notice on this branch is that HDC packets from the "enhanced" stream are no longer reported via `nrsc5_report_hdc`. (AM broadcasts often have an enhanced stream.)

That said, I see that the `stream_id` field is not exposed in `NRSC5_EVENT_HDC` events, so API clients wouldn't even have a useful way to separate the core stream from the enhanced stream! Nor is any sequencing information exposed, so an API client couldn't know when there's a gap (and therefore, that a decoder would need to be reset). I guess we should address these oversights in a separate PR.

argilo:
> One difference I notice on this branch is that HDC packets from the "enhanced" stream are no longer reported via nrsc5_report_hdc.

After thinking about this some more, I think it's actually a good thing to skip reporting enhanced packets. Ultimately, I think we should put them in a separate `enhanced_data` field alongside the `data` field in the `NRSC5_EVENT_HDC` event. We can leave that to a follow-up PR.

src/main.c
eean: they're all exactly 15536 or 15534 for me

TheDaChicken: The sample size is 15534 or 15536. The actual memory size is 15536 * sizeof(int16_t) which is equal to around 31072

argilo:
Thanks again for your work on this @TheDaChicken. I used your proposed changes here as inspiration, and made a simplified version in #399 which seems to work well. As long as no serious problems come up, I'll get that merged within a few days so we can finally have elastic buffering.

argilo:
I just merged #399, so I'll close this off.

We may return to the idea of maintaining an (almost) exact ratio of input samples to output samples in a future PR, to support blending (#80).

PR #344, opened 2024-05-05
Add MA3 (all digital) service mode

### LINKED ISSUE #240: Implement AM all-digital mode (MA3)
The AM receiver only supports the hybrid mode (MA1) currently. At least one station (WWFD in Frederick, Maryland) is using the all-digital mode (MA3), so it would be nice to be able to receive it. It would be very helpful to have an I/Q recording of this station.

argilo:
Fixes #240.

Here I have implemented support for MA3, the all-digital AM service mode. I tested that it is able to receive WWFD, using a recording made with W3HFU's KiwiSDR. As with the MA1 hybrid mode, only the "core" audio stream is decoded, and the "enhanced" stream is discarded. (Improvements to the HDC decoder would be needed to decode the enhanced stream. See #245 for a previous attempt at that.) I was able to decode WWFD's station logo:

7364_SLWWFD$010001

While working on this, I found a bug in gr-nrsc5's header spread calculations, which I fixed in https://github.com/argilo/gr-nrsc5/commit/c8b1b0f3deefc74ea628b683701909e03127ede6. After that change, I was also able to receive a synthesized MA3 signal generated by gr-nrsc5.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -204,7 +204,7 @@ void acquire_process(acquire_t *st)
 
             if (st->input->sync_state != SYNC_STATE_FINE)
             {
-                for (j = CENTER_AM - PIDS_2_INDEX_AM; j <= CENTER_AM + PIDS_2_INDEX_AM; j++)
+                for (j = CENTER_AM - PIDS_OUTER_INDEX_AM; j <= CENTER_AM + PIDS_OUTER_INDEX_AM; j++)
                 {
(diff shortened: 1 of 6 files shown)


argilo:
WWFD is broadcasting album art as well, but unfortunately I wasn't able to get any complete files because the bit error rate in the P3 logical channel is too high and some of the file chunks are lost.

I don't know yet whether the SNR of my recording is too low, or nrsc5's AM receiver is suboptimal.

Issue #346, opened 2024-06-20
API Improvements: Document Requirements / Add Missing Constants


TheDaChicken:
In the API, 

`nrsc5_pipe_samples_cs16` assumes input went through the half-band filter. This isn't documented as a requirement/function / is not labeled to require in the function name. 

I encountered a problem using 16 bits as a constant type for piping data from different SDRs which required the half-band filter. I have to implement a half-band filter to pipe that type.  

API does not include a few constants such as which may be required:
- IQ sample rate 
- audio sample rate
- audio sample count (provided in callbacks, not *really* needed)
- audio channel count (not very important)

Silly small tidbit: 
``__declspec(dllexport)`` is meant to be used in the header file.

argilo:
> `nrsc5_pipe_samples_cs16` assumes input went through the half-band filter.

Actually there is no requirement to use a half-band filter. The only requirement is that the input sample rate is 744187.5. I agree that it would be useful to add the required sample rates to the API documentation (1488375 for `nrsc5_pipe_samples_cu8`, 744187.5 for `nrsc5_pipe_samples_cs16`).

We could also document that the audio sample rate is always 44100.

argilo:
The reason for 744187.5 is that it is the most convenient sample rate for OFDM demodulation. Unfortunately, the RTL-SDR does not natively support this sample rate (it allows only 225000-300000 and 900000-3200000). Because of this limitation, nrsc5 samples at 744187.5 * 2 = 1488375 and uses a decimating half-band filter to reach the desired sample rate. The API allows samples to be injected either before or after nrsc5's filter. If samples are injected after the filter, the application can use whatever means it likes to achieve the correct sample rate.

TheDaChicken:
> Actually there is no requirement to use a half-band filter. The only requirement is that the input sample rate is 744187.5. I agree that it would be useful to add the required sample rates to the API documentation (1488375 for `nrsc5_pipe_samples_cu8`, 744187.5 for `nrsc5_pipe_samples_cs16`).

Thanks for letting me know. *my oblivious nature is a good example that it should probably be documented* 

> We could also document that the audio sample rate is always 44100.

Would be very useful. I have constants I added in my code that would make sense to be added.

TheDaChicken:
Should i make pull request for this?

argilo:
Yes, that would be useful.

pclov3r:
[...]

PR #347, opened 2024-06-25
Documents improvements

TheDaChicken:
Deals with #346 

It wasn't straightforward what the sample rate was meant to be for either of the 2 pipe functions. This adds macro and documents it for each function. 

~~This also deals with a warning that Clion IDE gave me. It warned me that `__declspec(dllexport)` was meant to be used in the header. If needed, this part can be removed. This is a nick-pick. CMake newer versions can do this automatically using `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)` though is not supported with this 2.8.~~

If you think of other constants that should be added, let me know. 

My program has a simple `#define NRSC5_MPS_PROGRAM 0x0` constant. Not sure if that is worth adding. It is simple zero.


support/nrsc5.py
```diff
@@ -29,6 +29,12 @@ class EventType(enum.Enum):
     PACKET = 13
 
 
+class SampleRates(float, enum.Enum):
+    CU8 = 1488375
+    CS16 = 744187.5
+    AUDIO = 44100
```
argilo, line 35: Grouping these into an `Enum` seems clumsy, since it takes extra work to get at the underlying values. Maybe it would be easier to make them contants:

```suggestion
SAMPLE_RATE_CU8 = 1488375
SAMPLE_RATE_CS16 = 744187.5
SAMPLE_RATE_AUDIO = 44100
```

argilo:
> This also deals with a warning that Clion IDE gave me. It warned me that `__declspec(dllexport)` was meant to be used in the header. If needed, this part can be removed.

If you don't mind, I'd rather consider that in a separate PR.

TheDaChicken:
> > This also deals with a warning that Clion IDE gave me. It warned me that `__declspec(dllexport)` was meant to be used in the header. If needed, this part can be removed.
> 
> If you don't mind, I'd rather consider that in a separate PR.

Alright, I created, #350. Other than the review, it should be all done!

src/defines.h
```diff
@@ -9,7 +9,7 @@
 #include <math.h>
 
 // Sample rate before decimation
-#define SAMPLE_RATE 1488375
+#define SAMPLE_RATE NRSC5_SAMPLE_RATE_CU8
```
argilo, line 12: Do you suppose we should get rid of `SAMPLE_RATE` and just use `NRSC5_SAMPLE_RATE_CU8` instead?

argilo (approved):
Looks good now. I'll merge once CI is happy.

TheDaChicken:
Also a few things before merge:

Are you sure that it decimates only 2 from cs16? What about AM? There is DECIMATION_FACTOR_AM in acquire.c and also looks like it decimates by 32 in nrsc5.c 

Should there be an NRSC5_PACKET_SAMPLES constant to prevent the 2048 magic number in main.c?

argilo:
> Are you sure that it decimates only 2 from cs16? What about AM? There is DECIMATION_FACTOR_AM in acquire.c and also looks like it decimates by 32 in nrsc5.c

That's a good point. For AM, there are five cascaded decimators, so the CS16 sample rate is 1488375 / 32 = 46511.71875. It would make sense to have separate constants for FM and AM.

> Should there be an NRSC5_PACKET_SAMPLES constant to prevent the 2048 magic number in main.c?

Yes, I think that would be nice. They are called audio frames in the specification, so perhaps the name could be `NRSC5_AUDIO_FRAME_SAMPLES`.

TheDaChicken:
> > Are you sure that it decimates only 2 from cs16? What about AM? There is DECIMATION_FACTOR_AM in acquire.c and also looks like it decimates by 32 in nrsc5.c
> 
> That's a good point. For AM, there are five cascaded decimators, so the CS16 sample rate is 1488375 / 32 = 46511.71875. It would make sense to have separate constants for FM and AM.
> 
> > Should there be an NRSC5_PACKET_SAMPLES constant to prevent the 2048 magic number in main.c?
> 
> Yes, I think that would be nice. They are called audio frames in the specification, so perhaps the name could be `NRSC5_AUDIO_FRAME_SAMPLES`.

Alright. Those have been added!

PR #348, opened 2024-07-14
Debian Build Configuration Files

luarvique:
Since nrsc5 is now successfully integrated with OpenWebRX+, I had to build it into a Debian package to make it easier for end users to install. This pull request contains the debian/ folder with build config files.

When merging, please, change my name to yours in control and changelog files. Also, assign it a proper version in changelog.


argilo:
@luarvique Are you prepared to maintain this? I don't really know anything about Debian packaging.

argilo:
> You obviously want to replace my name/address with yours, to begin with.

I don't want to put my name on something that I'm not prepared to maintain. That's why I asked whether you will maintain this.

> Then, every time you think it is worth a version increment, make a new entry in the changelog with the summary of changes, copypasting the previous entry for formatting.

Are you willing to do this?

luarvique:
Well, I will still package and host it in OWRX+ repos, so that OWRX users could easily install it. But doing it upstream (i.e. by yourself) will make it easier for non-OWRX users to install and use NRSC5. Especially if the package goes into main Linux repositories.

argilo:
OK. I'm not opposed to upstreaming the packaging files, as long as someone commits to keeping them up to date.

PR #350, opened 2024-07-19
Move DLLMarkup to header 

TheDaChicken:
This deals with a warning that Clion IDE gave me. 

`__declspec(dllexport)` is meant to be used in the header. 

Although CMake's newer versions can do this automatically using `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)`, we are still on CMake version 2.8. There is already a *DLLMarkup* so it should be okay.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -51,6 +51,16 @@
 #define NRSC5_SAMPLE_RATE_CS16_AM  46511.71875
 #define NRSC5_SAMPLE_RATE_AUDIO    44100
 
+#ifdef NRSC5_EXPORTS
+#ifdef __MINGW32__
+#define NRSC5_API __declspec(dllexport)
+#else
+#define NRSC5_API
+#endif
+#else
+#define NRSC5_API
+#endif
+
 enum
 {
     NRSC5_MODE_FM,
@@ -388,7 +398,7 @@ typedef struct nrsc5_t nrsc5_t;
  * @param[out] version character pointer that will reference the version string.
  * @return Nothing is returned
  */
-void nrsc5_get_version(const char **version);
+NRSC5_API void nrsc5_get_version(const char **version);
 
 /**
  * Retrieves a string corresponding to a service data type.
@@ -399,7 +409,7 @@ void nrsc5_get_version(const char **version);
  * This name will be quite short, e.g. "News" or "Weather". If the type is
  * not recognized, it will the string "Unknown".
  */
-void nrsc5_service_data_type_name(unsigned int type, const char **name);
+NRSC5_API void nrsc5_service_data_type_name(unsigned int type, const char **name);
 
 /**
  * Retrieves a string corresponding to a program type.
@@ -410,7 +420,7 @@ void nrsc5_service_data_type_name(unsigned int type, const char **name);
  * This name will be quite short, e.g. "News" or "Rock". If the type is
  * not recognized, it will the string "Unknown".
  */
-void nrsc5_program_type_name(unsigned int type, const char **name);
+NRSC5_API void nrsc5_program_type_name(unsigned int type, const char **name);
 
 /**
  * Initializes a session for a particular RTLSDR radio dongle.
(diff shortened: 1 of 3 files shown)


argilo:
> Although CMake's newer versions can do this automatically using set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON), we are still on CMake version 2.8.

`WINDOWS_EXPORT_ALL_SYMBOLS` was added in CMake 3.4, which was released in December 2018. I think that's long enough ago that we can safely bump the required CMake version.

argilo:
Pardon my ignorance, but what does this actually do? What is "a DLLMarkup"?

TheDaChicken:
> Pardon my ignorance, but what does this actually do? What is "a DLLMarkup"?

Good question. I even had to do a little research myself. I got the word DLLMarkup from CMake documentation. I am a little unsure too.

There are 2 ways a function can be exported to a DLL (or exported for public consumption):

1. DllExport markup or (`__declspec(dllexport)`) marked on a function. 
The correct proper full way seems to be something like this: https://gcc.gnu.org/wiki/Visibility#How_to_use_the_new_C.2B-.2B-_visibility_support 
or https://stackoverflow.com/questions/2164827/explicitly-exporting-shared-library-functions-in-linux
2. Provide a .def file 
3. There's a new way with `WINDOWS_EXPORT_ALL_SYMBOLS`

I went back to previous commits to see what happened. The linker option `--exclude-all-symbols` was added for Windows complication and then the markup was added. Other Systems were provided a map or sym file instead. A .def wasn't made for Windows to provide function names instead. This was all during commit: 044a92c23ca6ccc1111b833828c7b39ee3386974 or (apart from #170) 

So technically it is possible to provide a .def file instead of marking functions or include a macro like this in the header file. It seems exporting all symbols was disabled. 

Then again this project is on a VERY old CMake version + only GCC works. I always have to make modifications to CMakeList.txt to get complications working in some way. Things for PRs...

TheDaChicken:
Maybe I should ask, should a .def file be created instead?

argilo:
I don't know a lot about this, but it does appear the NRSC5_API stuff was meant to go in the header file instead of the implementation, so we should probably just move it there and keep things otherwise as before.

TheDaChicken:
> I don't know a lot about this, but it does appear the NRSC5_API stuff was meant to go in the header file instead of the implementation, so we should probably just move it there and keep things otherwise as before.

Yeah completely understandable! My apologies, if my explication wasn't good. I didn't even know the other opposites until you asked that question. 

If a provided .def file was made, it would match other operating systems without needing a macro. I don't know why that was chosen in the first place. ツ

include/nrsc5.h
```diff
@@ -51,6 +51,16 @@
 #define NRSC5_SAMPLE_RATE_CS16_AM  46511.71875
 #define NRSC5_SAMPLE_RATE_AUDIO    44100
 
+#ifdef NRSC5_EXPORTS
```
argilo, line 54: Is the `NRSC5_EXPORTS` conditional useful for something? It wasn't there before.

TheDaChicken: It's to make sure dllexport is only used when building the shared library (where export is needed). dllexport isn't needed when including the header normally

argilo: Would dllexport simply be ignored if building something other than a DLL?

TheDaChicken: In the examples I've seen, they created a macro to determine when dllexport is used. This is most likely because `__declspec(dllimport)` should be used instead when dll isn't being built (which I can add if you'd rather). 
There is normally a huge macro to handle other platforms. Other platforms have a file instead to provide these functions.

TheDaChicken: Am trying to be careful to not include random stuff that shouldn't be in as a public header.

TheDaChicken: bump

argilo: I guess that's fine, it seems like it shouldn't hurt anything.

PR #354, opened 2025-01-01
Update to latest FAAD2 Commit

TheDaChicken:
I saw FAAD2 now provides a CMakeLists which makes compiling a patched faad hdc a bit easier. I updated the patch for the latest commit. It's still not good as fdk-aac.

Works just fine for me here: https://youtu.be/KYT2bGjVlQo

I wanted to nrsc5 to be able to compile easier. Most people use CMake. I am not very familiar with libtoolize. 

I originally wanted to compile nrsc5 with MSVC because it doesn't depend on libc++. Plus, some people online disagree that -static is a good flag bla bla. nrsc5 requires lots of changes to get MSVC. I ended up just bringing libc with me.


argilo:
Stereo seems to be broken for secondary streams (HD2, HD3, etc). We'll have to investigate why that happened.

argilo:
I bisected faad2, and found that the problem began in this commit: https://github.com/knik0/faad2/commit/d5104df0b5b79431d9fcf90a0fe70857ddeec714

It added an extra check for parametric stereo in the DRM case, and I suspect we'll need to update our patch to do the same in the HDC case. (At present, parametric stereo is ignored completely for HDC.)

argilo:
I pushed an updated patch that incorporates the upstream change.

```diff
@@ -888,7 +902,11 @@ static uint16_t sbr_extension(bitfile *ld, sbr_info *sbr,
 #ifdef DRM_PS
     case DRM_PARAMETRIC_STEREO:
         /* If not expected then only decode but do not expose. */
+#ifdef HDC
+        if (sbr->Is_DRM_SBR || sbr->Is_HDC_SBR)
+#else
         if (sbr->Is_DRM_SBR)
+#endif
         {
             sbr->ps_used = 1;
         }
```

After this change, the audio output is unchanged (compared to nrsc5's master branch), apart from variations in the least significant bit.

PR #366, opened 2025-02-26
Make open and close functions thread-safe

### LINKED ISSUE #349: nrsc5_open_* API functions are not thread-safe
When trying to use nrsc5 library from OpenWebRX+ on the amd64/x64 architecture, things immediately crash with the following message:

`double free or corruption (!prev)`

Both arm64 and armhf architectures are ok. Also, command-line nrsc5 utility appears to be ok.

I have been trying to debug the issue but failed: the process gets killed before GDB has a chance to look at it. Tried MALLOC_CHECK_=2, did not help.

argilo:
Fixes #349.

The `nrsc5_open()`, `nrsc5_open_file()`, `nrsc5_open_pipe()`, `nrsc5_open_rtltcp()`, and `nrsc5_close()` API functions call into FFTW functions which are not thread-safe. (See https://www.fftw.org/doc/Thread-safety.html for an explanation of thread safety in FFTW.)

To make nrsc5's API functions thread-safe, I have protected the affected FFTW calls with a mutex. After this change, I was able to initialize and run many instances of nrsc5 in parallel from Python without any crashes.

@luarvique I hope this change will make it simpler for OpenWebRX+ to use nrcs5.

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -19,6 +19,7 @@
 #include "acquire.h"
 #include "defines.h"
 #include "input.h"
+#include "private.h"
 
 #define FILTER_DELAY 15
 #define DECIMATION_FACTOR_FM 2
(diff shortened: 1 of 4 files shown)


luarvique:
Thanks, based on my experience with FFTW, it should do the job!

PR #370, opened 2025-03-02
Target peak amplitude of -6 dBFS in automatic gain selection

### LINKED ISSUE #238: Implement automatic gain selection for AM
Currently automatic gain selection only works properly for FM. It should support AM as well, or at least display an error message indicating that it is not implemented and gain must be manually specified.

### LINKED ISSUE #333: Trouble with nrsc5 - Linux
Trouble with nrsc5 - Linux 

Have been experimenting with nrsc5 and a RTL-SDR blog V3 dongle. Recently upgraded my aerial which is working well for analog reception.  

Built and running nrsc5 on a Linux Mint Victoria 21.2, Cinnamon 5.8.4 tk, Kernel 6.2.0-36-generic x86_64 machine. [nrsc5 revision 80ab64e]

Decoding the nrsc-5 data is extremely sensitive to gain adjustments both manual and automatic. Not much in feedback from the software relating to gain or signal strength. 

Experiencing the loss of synchronization, odd audio artifacts and drop outs. MER can be all over the map. Typical data follows ...

11:10:54 MER: -5.5 dB (lower), -16.0 dB (upper)
11:10:54 BER: 0.090398, avg: 0.069607, min: 0.041372, max: 0.090398
11:10:54 Lost synchronization

Requesting suggestions resolving these types of problems. 

Thanks in advance,

argilo:
Fixes #238.
Fixes #333.

nrsc5's automatic gain selection currently works by stepping through all possible gain values, and measuring the amount of power present in the frequency ranges occupied by the HD Radio sidebands compared to the amount of power present at frequencies just below and just above the HD Radio sidebands.

This approach has some problems:

* It often chooses a gain value that nearly saturates the ADC, leaving little headroom in case the signal amplitude increases.
* The SNR measurement performs badly when adjacent-channel interference is present.
* The SNR measurement performs badly when the gain is high enough for clipping to occur.

I think it would be better to instead measure the peak amplitude of the incoming signal, and target -6 dBFS (i.e. 6 dB below full scale, where clipping would begin). So far this approach seems to work well, and selects more sensible gain values. When no interfering signals are present, the -6 dBFS signal level is more than sufficient to receive an error-free signal, while leaving a decent amount of headroom. (Bit errors only begin occurring when the signal level is below -24 dBFS.)

I'd appreciate feedback from anyone who can take this for a spin to confirm whether it works better than the old approach.

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -67,52 +67,6 @@ void input_set_skip(input_t *st, unsigned int skip)
     st->skip += skip;
 }
 
-static void measure_snr(input_t *st, const uint8_t *buf, uint32_t len)
-{
-    unsigned int i, j;
-
-    // use a small FFT to calculate magnitude of frequency ranges
-    for (j = 0; j + SNR_FFT_LEN <= len / 2; j += SNR_FFT_LEN)
-    {
-        for (i = 0; i < SNR_FFT_LEN; i++)
-            st->snr_fft_in[i] = CMPLXF(U8_F(buf[(i+j) * 2]), U8_F(buf[(i+j) * 2 + 1])) * pow(sinf(M_PI*i/(SNR_FFT_LEN-1)), 2);
-        fftwf_execute(st->snr_fft);
-        fftshift(st->snr_fft_out, SNR_FFT_LEN);
-
-        for (i = 0; i < SNR_FFT_LEN; i++)
-            st->snr_power[i] += normf(st->snr_fft_out[i]);
-        st->snr_cnt++;
-    }
-
-    if (st->snr_cnt >= SNR_FFT_COUNT)
-    {
-        // noise bands are the frequncies near our signal
-        float noise = 0;
-        for (i = SNR_NOISE_START; i < SNR_NOISE_START + SNR_NOISE_LEN; i++)
-        {
-            noise += st->snr_power[i];
-            noise += st->snr_power[SNR_FFT_LEN - i];
-        }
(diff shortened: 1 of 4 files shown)


argilo:
For anyone testing this, I'd be interested in any case where you're able to achieve a clearly higher MER (or clearly lower BER) by choosing your own gain instead of letting the automatic gain selection choose it.

pclov3r:
From testing a few stations arounds here appear to be getting the same MER at lower gain values.

Old method was setting around 32 dB gain and new method is setting around 28 dB

argilo:
> From testing a few stations arounds here appear to be getting the same MER at lower gain values.

It sounds like it's working as intended then. The previous algorithm could set the gain dangerously close to saturating the ADC, while the new algorithm always keeps it at least 6 dB below the saturation point.

markjfine:
Late to the party but I just got though playing with the latest master that has this included.
As I'm in a marginal signal area using an amplified TV antenna on the roof, this is working real well. Have noticed the gain figures that it settles on hits the manual ones I've used in the past. Not only does this avoid saturation, but it also helps temper adjacent channel splatter. Real good enhancement.

argilo:
Thanks for testing this @markjfine!

I also had a look over the library of recordings that I made prior to this change, and I see that the majority of them have at least a small amount of clipping, so clearly the old gain selection often set the gain too high.

If necessary, we could increase or decrease the -6 dB threshold in the future, but so far this value seems to work well.

Issue #293, opened 2022-12-27
Error No rule to make target 'faad2-prefix/lib/libfaad_hdc.a',


emrecio:
I get through most of the compile process but fail at last step...

[OpenSuSE 15.4...]

╭─erecio@moz /usr/local/src/RTL/nrsc5/foo ‹master●› 
╰─$ make -k                                                                                                            2 ↵
[ 28%] Built target faad2_external
[ 85%] Built target nrsc5_object
[ 89%] Built target nrsc5_static
make[2]: *** No rule to make target 'faad2-prefix/lib/libfaad_hdc.a', needed by 'src/nrsc5'.
make[2]: Target 'src/CMakeFiles/app.dir/build' not remade because of errors.
make[1]: *** [CMakeFiles/Makefile2:133: src/CMakeFiles/app.dir/all] Error 2
make[2]: *** No rule to make target 'faad2-prefix/lib/libfaad_hdc.a', needed by 'src/libnrsc5.so'.
make[2]: Target 'src/CMakeFiles/nrsc5.dir/build' not remade because of errors.
make[1]: *** [CMakeFiles/Makefile2:187: src/CMakeFiles/nrsc5.dir/all] Error 2
make[1]: Target 'all' not remade because of errors.
make: *** [Makefile:136: all] Error 2
make: Target 'default_target' not remade because of errors.

Karlomanio:
Is there any workaround for this issue?

whelanh:
I have the same error trying to compile on Opensuse Tumbleweed

argilo:
I just built nrsc5 in an opensuse/tumbleweed Docker container and didn't have any issue.

Do you have all the dependencies installed? I needed `libao-devel fftw3-devel rtl-sdr-devel libusb-1_0-devel`.

argilo:
Here are the commands I used to install dependencies:

```
zypper install -t pattern devel_C_C++
zypper install git cmake libao-devel fftw3-devel rtl-sdr-devel libusb-1_0-devel
```

After that, following the rest of the build instructions in the readme worked fine.

argilo:
I'll close this off since I haven't received a response, and building appears to work correctly on openSUSE.

argilo:
I have also added openSUSE build instructions to the readme.

PR #371, opened 2025-03-06
Fix build on openSUSE 15.6

eean:
autoconf wants to install things into lib64 not lib by default on openSUSE

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,7 +226,7 @@ if (USE_FAAD2)
         PATCH_COMMAND patch -p1 -Ni "${CMAKE_SOURCE_DIR}/support/faad2-hdc-support.patch" || exit 0
         COMMAND sh ./bootstrap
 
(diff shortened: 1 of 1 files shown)


argilo:
Is this the same issue as #293?

eean:
> I tried to reproduce this in an openSUSE container but didn't have any luck. Any idea why that might be?

Yeah it's a difference between 15.6 and Tumbleweed it appears. I just created a Tumbleweed distrobox, followed those directions, and was able to build without my change.

argilo:
Ah.

I suppose this change makes sense, since `/lib/` is specified here:

https://github.com/theori-io/nrsc5/blob/3b71ec72bc2b809f81f6e65b68e495432545c83d/CMakeLists.txt#L233

argilo:
FYI, this change was short-lived because #354 switched to building FAAD2 with CMake, and I just merged that change.

@eean Would you mind checking whether the build process still works on your openSUSE machine?

eean:
@argilo oh I've been pulling from master and building since then and haven't noticed anything. Glad to see cmake continue to take over.

argilo:
Thanks for confirming. :+1:

PR #375, opened 2025-03-18
Implement COMM Tag

### LINKED ISSUE #352: COMM Tag
I wanted to post that I found a station over here (97.3 KIRO) that is using the COMM tag. 

image
image

I don't know how useful the comment tag is. Seems it's just a simple Language field, a short description field, and a content field providing a description.

TheDaChicken:
Fixes #352 

KIRO-FM here in Seattle uses the only unimplemented id3 tag in nrsc5. 

This tag in use:
```
20:52:03 BER: 0.000000, avg: 0.000000, min: 0.000000, max: 0.000000
20:52:04 Audio bit rate: 19.3 kbps
20:52:05 MER: 13.9 dB (lower), 14.0 dB (upper)
20:52:05 BER: 0.000000, avg: 0.000000, min: 0.000000, max: 0.000000
20:52:05 Title: Seattle Sports
20:52:05 Artist: Brought to you by Quantum Fiber
20:52:05 Album: Seattle Sports
20:52:05 Genre: Sports
20:52:05 Comment: lang=eng Comm https://www.quantumfiber.com
20:52:06 MER: 13.9 dB (lower), 14.0 dB (upper)
20:52:06 BER: 0.000000, avg: 0.000000, min: 0.000000, max: 0.000000
```


include/nrsc5.h
```diff
@@ -263,6 +263,22 @@ struct nrsc5_sis_dsd_t
...
+struct nrsc5_id3_comm {
+    struct nrsc5_id3_comm *next; /**< Pointer to next element or NULL */
+    char *lang; /**< language code, e.g. "eng" */
+    char *short_content; /**< short content description */
+    char *actual_text; /**< actual text */
+};
+/**
+ * Defines a typename for struct nrsc5_id3_comm_tag
+ */
+typedef struct nrsc5_id3_comm nrsc5_id3_comm;
```
argilo, line 280: All other structs have a `_t` suffix, so we should probably use that here for consistency.

Also, it might be clearer to use the full `comment` instead of just `comm`.

include/nrsc5.h
```diff
@@ -320,6 +336,7 @@ struct nrsc5_event_t
             const char *artist;
             const char *album;
             const char *genre;
+            nrsc5_id3_comm *comments;
```
argilo, line 339: This should probably go at the end of the struct. Inserting it in the middle will break ABI compatibility.

src/output.c
```diff
@@ -239,6 +246,34 @@ static void output_id3(output_t *st, unsigned int program, uint8_t *buf, unsigne
...
         }
+        else if (memcmp(tag, "COMM", 4) == 0)
+        {
+            if (frame_len < 5)
+            {
+                log_warn("bad COMM tag (frame_len %d)", frame_len);
+            }
+            else
+            {
+                uint8_t *delim = memchr(data+4, 0, frame_len);
```
argilo, line 257: I don't think this will work for UCS2, where the terminator is `00 00`.

Also, it appears this could read past the end of the buffer; `frame_len` should be `frame_len - 4`.

support/cli.py
```diff
@@ -248,6 +248,8 @@ def callback(self, evt_type, evt):
                     logging.info("Genre: %s", evt.genre)
                 if evt.ufid:
                     logging.info("Unique file identifier: %s %s", evt.ufid.owner, evt.ufid.id)
+                for comment in evt.comments:
+                    logging.info("Comment: lang=%s %s %s", comment.lang, comment.short_content, comment.actual_text)
```
argilo, line 252: This should probably go between Genre and UFID to match the C CLI.

support/nrsc5.py
```diff
@@ -135,9 +135,10 @@ class ProgramType(enum.Enum):
 BER = collections.namedtuple("BER", ["cber"])
 HDC = collections.namedtuple("HDC", ["program", "data"])
 Audio = collections.namedtuple("Audio", ["program", "data"])
+COMM = collections.namedtuple("COMM", ["lang", "short_content", "actual_text"])
```
argilo, line 138: As with the C struct, maybe this would be more understandable as `Comment`?

src/output.c
```diff
@@ -239,6 +263,34 @@ static void output_id3(output_t *st, unsigned int program, uint8_t *buf, unsigne
...
+                uint8_t *end = data + frame_len;
+
+                if (delim)
+                {
+                    nrsc5_id3_comment_t* prev = comm;
+
+                    comm = calloc(1, sizeof(nrsc5_id3_comment_t));
+                    comm->lang = strndup((char*) data + 1, 3);
+                    comm->short_content = id3_encode_utf8(enc, data + 4, delim - data - 4);
+                    comm->actual_text = id3_encode_utf8(enc, delim + 1, end - delim);
```
argilo, line 285: Two bugs here:

1. The starting point for the string is incorrect in the UCS-2 case; it should be `delim + 2`.
2. The length calculation is incorrect because it doesn't subtract off the length of the delimiter itself (1 or 2). As a result, this reads past the end of the data buffer.

Perhaps it would be simpler to change `id3_strlen` to something like `id3_next_string` and have it return a pointer to the first byte after the delimiter.

argilo: On second thought, that approach would just move the problem and make it harder to calculate the length of `short_content`.

argilo:
I fixed the buffer overrun and renamed some fields to hopefully be a bit clearer. Let me know if you're happy with this now, and I'll get it merged.

Decoding appears to work correctly with the couple of KIRO recordings you sent me.

TheDaChicken:
Due to a bug with ID3 frame length in nrsc5. (@argilo fixed it).

I kept overrunning without realizing since frame length was wrong. 

This is actually the full tag: 
```
Local time zone offset: -563 minutes, DST sched. 1, local DST? no, regional DST? no
17:28:37 BER: 0.000000, avg: 0.000203, min: 0.000000, max: 0.000610
17:28:37 Title: Seattle Sports
17:28:37 Artist: Brought to you by Quantum Fiber
17:28:37 Album: Seattle Sports
17:28:37 Genre: Sports
17:28:37 Comment: lang=eng Comm https://www.quantumfiber.com/?gad_source=1&gclid=Cj0KCQiAvvO7BhC-ARIsAGFyToUg6cWv3odVk87b03c-C-8wcBatJzAg1Qznn5Z8kgnBo10ninc9i8AaAkI-EALw_wcB&gclsrc=aw.ds
17:28:37 XHDR: 0 BE4B7536 27500
17:28:37 Audio bit rate: 18.8 kbps
17:28:37 MER: 13.7 dB (lower), 13.9 dB (upper)
```

PR #379, opened 2025-03-22
Feed soft decisions into Viterbi decoder

argilo:
I recently compared nrsc5's low-SNR performance against a Sangean HDT-20, and found that the HDT-20 is able to go about 2 dB lower than nrsc5. The reason for the different seems to be that nrsc5 uses hard-decision decoding:

https://github.com/theori-io/nrsc5/blob/ddd097988b45fedaf5ff996c18a8e7e0c08ce1da/src/sync.c#L455

After switching to soft-decision decoding, nrsc5 appears to be on par with the HDT-20, and reception is reliable (i.e. no invalid HDC packets) all the way down to 4 dB SNR.

### DIFF
--- a/src/sync.c
+++ b/src/sync.c
@@ -69,6 +69,12 @@ static uint8_t gray8(float f)
         return 1;
 }
 
+static inline int8_t demod(float x, float mult)
+{
+    float clamped = fmaxf(fminf(x, 1), -1);
+    return lroundf(clamped * mult);
+}
+
 static uint8_t qpsk(complex float cf)
 {
     return (crealf(cf) < 0 ? 0 : 1) | (cimagf(cf) < 0 ? 0 : 2);
@@ -452,7 +458,6 @@ void sync_process_fm(sync_t *st)
         float mult_lb = fmaxf(fminf(mer_lb * 10, 127), 1);
         float mult_ub = fmaxf(fminf(mer_ub * 10, 127), 1);
 
-#define DEMOD(x) ((x) >= 0 ? 1 : -1)
         for (int n = 0; n < BLKSZ; n++)
         {
             float complex c;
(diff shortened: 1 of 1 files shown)


src/sync.c
```diff
@@ -69,6 +69,12 @@ static uint8_t gray8(float f)
         return 1;
 }
 
+static inline uint8_t demod(float x, float mult)
+{
+    float clamped = fmaxf(fminf(x, 1), -1);
```
argilo, line 74: Clamping the samples like this reduces performance slightly in low-SNR AWGN channels. But it noticeably improves performance in cases where there is adjacent-channel interference.

I believe the issue is that the noise power estimate (which is computed once for the entire sideband) is inaccurate when narrow-band interference corrupts only a small part of the sideband. Perhaps in the future we could switch to more granular noise power estimates.

src/sync.c
```diff
@@ -69,6 +69,12 @@ static uint8_t gray8(float f)
...
+    return lroundf(clamped * mult);
```
argilo, line 75: Explicit rounding is import here, to avoid round-toward-zero distortion.

argilo:
I tested this against my library of I/Q recordings, and it appears to make an across-the-board improvement. High-SNR recordings decode as before, and low-SNR recordings have many more audio packets decoded.

markjfine:
Did some initial testing on fringe stations shortly after this was committed. I seemed to have fewer disruptions on marginal stations and improved syncing on the usual really weak ones. My eyes always gravitate towards the BER in these cases, and I'd venture to say there's about 2dB more headroom on average now.

argilo:
Thanks for testing. I'm happy to hear you observed a similar improvement in low-SNR performance (just under 2 dB, as far as I can tell).

Issue #380, opened 2025-03-25
PIDS decoding starts late


argilo:
The PIDS logical channel uses a short (1-block = ~93 ms) interleaver so that a receiver can quickly retrieve SIS metadata such as the station's call sign. But nrsc5 does not transition into the `SYNC_STATE_FINE` state and begin decoding data until block 0 of a 16-block frame is detected, thereby adding as many as 15 blocks (= ~1.4 seconds) of extra delay until SIS is finally decoded:

https://github.com/theori-io/nrsc5/blob/660630f3f6d654083c711f76a632ee6d73c9353b/src/sync.c#L351-L368

It would be better to begin decoding data as soon as any frame is detected. This would make a scan feature (#103) more practical, and could also help with receiving sporadic signals.

PR #382, opened 2025-03-25
Use 32 symbols for coarse acquisition

argilo:
Currently nrsc5 uses two blocks (= 64 symbols) for coarse (i.e. pre-FFT) OFDM synchronization. This seems to be overkill, as the detected correlation peak remains strong even at 0 dB SNR where the data carried in QPSK subcarriers is not decodable. Switching to a single block (= 32 symbols) has negligible impact on the number of decoded audio packets. AM reception at low SNR suffers slightly, but only because it uses a poor design. (It should switch to using the AM carrier and reference subcarriers for synchronization, as intended.)

This change also removes an obstacle to implementing elastic buffering (#343): because the "acquire" stage presently operates two blocks at a time, and because the pair of blocks is not necessarily aligned with the start of a frame, one extra block of latency can occur some of the time.

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -19,7 +19,7 @@
 // OFDM symbols per L1 block
 #define BLKSZ 32
 // symbols processed by each invocation of acquire_process
-#define ACQUIRE_SYMBOLS (BLKSZ * 2)
+#define ACQUIRE_SYMBOLS (BLKSZ)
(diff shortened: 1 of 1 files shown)


argilo:
Reducing the amount of buffering in the "acquire" stage will also make it possible to decode PIDS earlier (#380).

PR #385, opened 2025-03-27
Speed up automatic gain selection with binary search

argilo:
In #370, automatic gain selection was changed to measure only the amplitude at the ADC. Since the relationship between gain and amplitude is monotonic, there's no reason to check every gain value, and we can speed things up with a binary search.

For the R820T tuner, no more than 5 gain values need to be checked (vs. 29 before).

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -39,11 +39,15 @@ static int do_auto_gain(nrsc5_t *st)
     if (gain_count < 0)
         goto error;
 
-    for (int i = 0; i < gain_count; i++)
-    {
-        int gain = gain_list[i];
+    int low = 0;
+    int high = gain_count - 1;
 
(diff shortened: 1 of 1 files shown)


src/nrsc5.c
```diff
@@ -69,7 +73,7 @@ static int do_auto_gain(nrsc5_t *st)
 
         uint8_t max_sample = 0;
         uint8_t min_sample = 255;
-        for (int j = 0; j < len; j++)
+        for (int j = len/4; j < len; j++)
```
argilo, line 76: I had to change this to ignore the samples towards the beginning of the buffer, since they are leftover from the "old" gain value. This wasn't an issue when we tested gains in increasing order, but the binary search adjusts the gain in both directions.

PR #386, opened 2025-03-28
Start decoding PIDS, P3, and P4 earlier

### LINKED ISSUE #380: PIDS decoding starts late

argilo:
Fixes #380.

Currently, nrsc5 doesn't decode data from any of the logical channels until block 0 of a 16-block transfer frame arrives. This can take up to 1.4 seconds.

PIDS data arrives one block at a time, so we can begin decoding it as soon as synchronization is achieved. SIS data appears about a second after starting nrsc5.

P3 and P4 data arrives two blocks at a time, so we can begin decoding it as soon as an even-numbered block arrives. Audio still takes a few seconds to begin playing, because these logical channels use an interleaver that is 32 blocks long.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -429,17 +429,40 @@ void decode_process_p1_p3_am(decode_t *st)
     }
 }
 
+void decode_set_block(decode_t *st, unsigned int bc)
+{
+    st->idx_pm = 720 * BLKSZ * bc;
+    if (bc == 0)
+        st->started_pm = 1;
+
+    st->interleaver_px1.idx = (st->interleaver_px1.length / 2) * (bc % 2);
+    st->interleaver_px2.idx = (st->interleaver_px2.length / 2) * (bc % 2);
+    if ((bc % 2) == 0)
+    {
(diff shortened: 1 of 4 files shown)


argilo:
Nothing changes for the P1 channel, because it arrives one transfer frame at a time, so there's no choice but to wait for the beginning of the next frame.

Issue #387, opened 2025-03-28
AM synchronization fails for MA3 (all-digital)


argilo:
Currently, coarse OFDM synchronization for AM stations works much like FM: it filters the signal to isolate the OFDM sidebands, then uses correlation to estimate the timing & frequency errors. The filter selects the MA1 primary sidebands (10-15 kHz from the carrier), but these sidebands are not present in MA3 signals, where the sidebands are less than 10 kHz from the carrier. As a result, coarse synchronization fails for MA3 signals, and synchronization is only achieved by accident after a long wait.

AM synchronization should instead use the AM carrier to estimate frequency error, and use the reference subcarriers (+/- 181.7 Hz) to estimate timing error. This technique would work equally well for MA1 and MA3.

Related issue: #241

PR #388, opened 2025-03-29
Move block synchronization to the acquire stage

argilo:
nrsc5 needs to maintain synchronization both with individual OFDM symbols, and with blocks (groups of 32 OFDM symbols). Currently the "acquire" stage takes care of the former. It has a 33-symbol buffer, and wants the first 32 of those symbols to be occupied by a block, and it carries over the 33rd symbol to the next invocation of `acquire_process`. But if the OFDM symbol alignment is off, it adjusts that carryover amount by up to +/- one half of a symbol, so that symbol alignment is restored.

https://github.com/theori-io/nrsc5/blob/4005811e2f2d95c1b37ff9f29d426f57daa3ae76/src/acquire.c#L257-L259

Meanwhile, when the "sync" stage finds which of the 32 OFDM symbols is the start of a block, it invokes `input_set_skip` to ask the "input" stage to discard that number of OFDM symbols, so that the next batch sent to the "acquire" stage will be aligned to the beginning of a block.

Unfortunately, this system causes a bug: at the time the "input" stage skips over some OFDM symbols, the "acquire" stage still has carryover samples in its buffer, and the rest of its buffer is then filled with samples from a later time. The discontinuity causes bit errors and disturbs the synchronization process.

Since the "acquire" stage is already capable of carrying over an arbitrary number of samples, we can make it responsible for block alignment. The "input" stage asks it to carry over enough additional symbols such that the first symbol will move to the beginning of its buffer.

I tested this with my collection of recordings, and the number of decoded packets increased (except for a couple MA3 recordings where synchronization is luck-of-the-draw due to #387).

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -254,9 +254,15 @@ void acquire_process(acquire_t *st)
         sync_push(&st->input->sync, st->fftout);
     }
 
-    keep = st->fftcp + (st->fftcp / 2 - samperr);
+    keep = st->fftcp + (st->fftcp / 2 - samperr) + st->keep_extra;
+    st->keep_extra = 0;
     memmove(&st->in_buffer[0], &st->in_buffer[st->idx - keep], sizeof(cint16_t) * keep);
     st->idx = keep;
+}
+
+void acquire_keep_extra(acquire_t *st, int extra)
+{
+    st->keep_extra = extra;
 }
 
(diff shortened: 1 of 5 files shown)


argilo:
A side benefit of this change is that since we now carry over symbols instead of discarding them, synchronization is achieved one block earlier.

PR #394, opened 2025-04-10
Handle Ctrl+C on Windows

### LINKED ISSUE #314: On Windows, NRSC5 should quit after reading an I/Q file
On Linux I've noticed NRSC5 quits after reading an I/Q file. 

This currently does not occur on Windows.

argilo:
Pressing Ctrl+C on Windows does nothing. This happens because the input thread uses `_getch()` to receive keyboard input, and this function receives Ctrl+C, preventing the (default) signal handler from being invoked.

Instead of using `_getch()` we can do as the non-Windows version does: disable console echo, and receive keystrokes by reading from stdin. After this change, Ctrl+C works as expected.

I had hoped this change might also resolve #314, but it seems that `pthread_cancel` does not interrupt the `read` function on Windows.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -513,7 +513,12 @@ static void *input_main(void *arg)
     if (!isatty(STDIN_FILENO))
         return NULL;
 
-#ifndef __MINGW32__
+#ifdef __MINGW32__
+    HANDLE hStdin = GetStdHandle(STD_INPUT_HANDLE);
+    DWORD mode = 0;
+    GetConsoleMode(hStdin, &mode);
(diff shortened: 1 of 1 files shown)


pclov3r:
This works great here

Thanks for fixing this.

argilo:
> Q doesn't work if you use a pipe but works normally with a file.

Yeah, I don't expect any of the keyboard commands (except Ctrl+C) will work if you're reading from stdin with something like `-r -` because the keyboard commands are also coming in via stdin.

It could be that worked before with `_getch()` (I didn't check). Are you happy with the new version? I like it better because it's one less difference between Windows & non-Windows.

pclov3r:
Yes, This needed to be fixed for awhile and glad it is now.

PR #399, opened 2025-05-03
Implement elastic buffering for audio output

### LINKED ISSUE #177: Audio events are bursty
Currently, audio events are reported as soon as audio packets are decoded. In the P1 channel, decoding happens one frame at a time, and frames are 1.486 seconds long. This places a heavy burden for buffering on the user of the API.

An alternative would be to have libnrsc5 take on some of the responsibility for buffering. It could store the decoded audio packets and report them gradually as new IQ samples arrive. In fact, this would be necessary if we ever wanted to implement analog blending (#80). The audio PDU header has latency and sequence number fields indicating exactly when audio frames should be played back in relation to the incoming signal.

### LINKED ISSUE #261: Python CLI has lots of underrun issues
Hi, so I decided to mess with the python cli and there is a problem. A pretty nasty one too.

This message will pop up
ALSA lib pcm.c:8526:(snd_pcm_recover) underrun occurred

The audio will become clickly and bad even when the signal is strong and the C program works fine. Can a python pro that is a part of this project help figure this out? It is also what is causing t[his issue on nrsc5-gui](https://github.com/cmnybo/nrsc5-gui/issues/17) so if the cli code could be fixed I can likely just use that in this program and boom! problem solved! So let's fix this!

### LINKED ISSUE #330: Underruns & cracking noises

argilo:
Replaces #343.
Fixes #177.
Fixes #261.
Fixes #330.

The NRSC-5 standard uses a variable-rate audio codec. To account for variations in the instantaneous bit rate, the standard specifies that both the transmitter and receiver use an elastic buffer, so that audio packets enter the transmitter and leave the receiver at a fixed rate, even though the number of packets sent in individual audio PDUs may vary. Unfortunately, the standard does not explain how elastic buffering is implemented, so we are left to guess many of the details.

Up until now, nrsc5 did not implement elastic buffering at all, and instead reported audio packets through its API immediately upon decoding. In this PR I have added elastic buffers (one per stream) which queue audio packets. A batch of audio packets (2 for FM, 4 for AM) is released each time an L1 block is decoded. (In a future PR, we could perhaps even shorten the cadence so that audio packets are released one at a time.)

API clients shouldn't need to make any changes to take advantage of elastic buffering. They will notice that audio packets are now produced in much smaller bursts (2 for FM, 4 for AM) and that the burst size is always identical. As a result, API clients should not need to buffer more than a handful of audio packets.

This PR is a simplified version of #343. That PR attempted to maintain an exact ratio of input RF samples to output audio samples, and I have dispensed with that requirement for now to simplify the implementation. We may need to return to that approach in the future to implement analog blending (#80).

### DIFF
--- a/src/acquire.c
+++ b/src/acquire.c
@@ -105,6 +105,8 @@ void acquire_process(acquire_t *st)
     if (st->idx != (unsigned int)st->fftcp * (ACQUIRE_SYMBOLS + 1))
         return;
 
+    output_advance(st->input->output);
+
     if (st->input->sync_state == SYNC_STATE_FINE)
     {
         samperr = st->fftcp / 2 + st->input->sync.samperr;
--- a/src/decode.c
+++ b/src/decode.c
@@ -391,41 +391,47 @@ void decode_process_pids_am(decode_t *st)
 
 void decode_process_p1_p3_am(decode_t *st)
 {
-    int total_errors = 0;
+    unsigned int block = st->idx_pu_pl_s_t / (PARTITION_WIDTH_AM * BLKSZ) - 1;
 
-    interleaver_ma1(st);
+    if (block == 0)
+        st->am_errors = 0;
 
-    if (st->am_diversity_wait > 0)
-    {
-        st->am_diversity_wait--;
-        return;
-    }
-
-    for (int block = 0; block < 8; block++)
+    if (st->am_diversity_wait == 0)
     {
         nrsc5_conv_decode_e1(st->viterbi_p1_am + (block * P1_FRAME_LEN_AM * 3), st->scrambler_p1_am, P1_FRAME_LEN_AM);
-        total_errors += bit_errors_p1_am(st->viterbi_p1_am + (block * P1_FRAME_LEN_AM * 3), st->scrambler_p1_am);
+        st->am_errors += bit_errors_p1_am(st->viterbi_p1_am + (block * P1_FRAME_LEN_AM * 3), st->scrambler_p1_am);
         descramble(st->scrambler_p1_am, P1_FRAME_LEN_AM);
         frame_push(&st->input->frame, st->scrambler_p1_am, P1_FRAME_LEN_AM, P1_LOGICAL_CHANNEL);
+
+        if (block == 7)
+        {
+            if (st->input->sync.psmi != SERVICE_MODE_MA3)
+            {
+                nrsc5_conv_decode_e2(st->viterbi_p3_am, st->scrambler_p3_am, P3_FRAME_LEN_MA1);
+                st->am_errors += bit_errors_p3_ma1(st->viterbi_p3_am, st->scrambler_p3_am);
+                descramble(st->scrambler_p3_am, P3_FRAME_LEN_MA1);
+                frame_push(&st->input->frame, st->scrambler_p3_am, P3_FRAME_LEN_MA1, P3_LOGICAL_CHANNEL);
+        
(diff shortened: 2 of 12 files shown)


argilo:
Some notable changes:

* The audio decoder is now reset any time an audio packet is missed. This should help reduce audio artifacts, since the decoder carries over state from one packet to the next, and does not cope well with omitted packets.
* When an audio packet is missed, the API will still report an audio event, but with silence in the audio buffer.
* The API now only reports HDC events for the "core" audio stream, and not the "enhanced" stream. (It was a mistake to report both, since the API does not provide any way to distinguish the two streams. In a future PR, we can add separate fields for enhanced packets, as proposed in #245).

argilo:
I'll leave this open for a couple days to allow some time for testing. Perhaps some of the usual suspects like @markjfine, @TheDaChicken, and @pclov3r might have some feedback.

markjfine:
Seems ok. Trick was finding an intermittent station to test on. The audio is either there or it isn't, with no artifacts when it transitions. Will also continue to test on these as well as stronger stations to see if any overflows happen over an extended listening period.

markjfine:
So while I'm testing, just noticed WWWT removed their broken HD5. 🤦🏻‍♂️

argilo:
Are they still using MP11, or have they switched back to MP3?

markjfine:
Lol. They've removed it completely. Back down to only 4 programs now.
Screenshot 2025-05-04 at 12 55 26 PM

markjfine:
11

markjfine:
Uploaded an iq file to https://fineware-swl.com/downloads/WWWT_2505041719.iq

argilo:
Got it, thanks!

The P4 logical channel is completely empty, so they're just wasting the bandwidth.

argilo:
More specifically, the logical channel contains an endless stream of `0x22 0x22 0x22 [...] 0x22`.

markjfine:
Ok, so I've had this thing going for hours now on WAMU. There were a few crc dropouts, but I'm in a fringe area and that's to be expected (someone could've moved their hay baler on the next road over). No artifacts at any time. 👍🏼

pclov3r:
Doing some testing it seems to be ok and match the behavior that would see on a real hd radio.

pclov3r:
Actually, It appears the elastic buffering doesn't do so well with low SNR conditions and fading.

https://gofile.io/d/UjiYi9

Without the elastic buffering it appears without it decodes much easier.

argilo:
It looks like the old & new produce the same amount of audio, it's just that the new version inserts silence where the missing packets are, whereas the old version doesn't.

PR #402, opened 2025-05-06
Report information from audio PDU headers

### LINKED ISSUE #374: Show the Blend Control mode setting
Though it's not commonly used it still might be useful if the blend control mode was reported.

### LINKED ISSUE #398: Provide programs via alternative way
Some HD stations like KNKX or WETS don't provide their multichannel programs in sis. 

This results in the client not being able to identify any programs unless audio packets or hdc packets were checked.

argilo:
Fixes #374.
Fixes #398.

Neither SIS audio descriptors nor SIG are mandatory, leaving applications with no reliable way to know of the existence of an audio program. Audio PDU headers are always present, and convey useful information about audio programs, so I think it makes sense to expose this information through the API.

Here I've added a new event (`NRSC5_EVENT_AUDIO_SERVICE`) which exposes ~a linked list of~ audio services. The following information is available:

* program number
* access (public / restricted)
* program type
* codec mode
* blend control
* digital audio gain
* digital audio delay
* codec latency

The C & Python API clients display this information on "Audio service" lines:

```
16:16:36 Audio service 0: PUBLIC, type: Adult Hits, codec: 0, blend: ENABLE, gain: 0 dB, delay: 96, latency: 8
16:16:36 Audio service 1: PUBLIC, type: Talk, codec: 0, blend: DISABLE, gain: 0 dB, delay: 0, latency: 8
16:16:36 Audio service 2: PUBLIC, type: Sports, codec: 13, blend: DISABLE, gain: 0 dB, delay: 0, latency: 8
```

It is unfortunate that audio program information is provided by three separate sources (SIS, PDU headers, SIG), but I think it's best for the API to expose the information in its raw form and let applications decide how best to process it.

SIS audio descriptors are decoded very quickly because they are part of the low-latency PIDS channel. But not all stations broadcast them. PDU headers take longer to appear, but always arrive before audio is decoded. SIG typically arrives last, since the data is contained in a file which takes time to download. But stations that have only a single audio program and no data services may not broadcast SIG.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -79,7 +79,7 @@ enum
     NRSC5_SIG_COMPONENT_DATA
 };
 
-/**  Represent a channel component.
+/**  Represents a service component.
  *
  * An element of a linked list that accompanies a nrsc5_sig_service_t
  * describing a component of a SIG record. This provides further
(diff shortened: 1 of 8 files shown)


argilo:
> Here I've added a new event (NRSC5_EVENT_AUDIO_SERVICES) which exposes a linked list of audio services.

I'm having second thoughts about this. I copied the linked-list approach over from SIS events. There it maybe makes more sense, since all SIS data is grouped into a single event. For PDU headers, we could always report them on an individual bases. The only downside is that there would be no way to indicate that a service has been removed. (But nrsc5 doesn't handle program removals, and it would probably be fine to require a station to be re-tuned after a program change anyway.)

argilo:
After a discussion with @TheDaChicken I decided that it's best to remove the linked list and report audio programs individually. I'll update the PR with that change soon.

argilo:
I removed the linked list. Things now look a bit better in the CLI applications, since each audio service is only reported once.

Issue #295, opened 2023-02-15
Feature Request: Dump only DWRO named image files


duracell80:
It would be very helpful to reduce the number of writes to an SSD drive when using the --dump-aas-files flag. Excluding station graphics, album art and traffic would really help clean up the directory. The weather radar images are updated roughly every 10 minutes OTA but get overwritten a few times in that time. Same for traffic, if there was a way to filter and dump only the files needed that would really help.

An example of the filename I see for radar:
1569_DWRO_******_rev03_20230215_2200_0621.png

Here I'm dumping the files and sending the audio output to null.

```
$DIR_BASE=/tmp/nrsc5
$DEV=0
$BAND=0
$FREQUENCY=107.5

nrsc5 -o - -d $DEV_IDEX -q --dump-aas-files $DIR_BASE $FREQUENCY $BAND > /dev/null 2>&1 &
```

I can then collect the overlays from the /tmp/nrsc5 directory into my home and into an animation video or gif with ffmpeg.

argilo:
Thanks for the suggestion.

I think instead of using filenames, it would be best to determine which data component a file belongs to using the MIME type information in the SIG table. For instance, here's a station that broadcasts weather maps (`TTN_STM_WEATHER = 0xEF042E96`) on port 0820:

```
18:21:12 SIG Service: type=data number=22 name=TTN STM
18:21:12   Data component: id=0 port=0810 service_data_type=66 type=3 mime=FF8422D7
18:21:12   Data component: id=2 port=0820 service_data_type=66 type=3 mime=EF042E96
```

Weather files can then be seen on port 0820:
```
18:21:12 LOT file: port=0810 lot=249 name=TMT_035iuu_1_2_20240612_1514_00f9.png size=5466 mime=4F328CA0 expiry=2024-06-12T15:45:00Z
```

Whether such a feature belongs in nrsc5's sample CLI applications or in external program is another question, but I can already see that it would be more complicated than necessary to implement in an application because the `NRSC5_EVENT_LOT` API event does not include the data component's MIME type (unlike the `NRSC5_EVENT_STREAM` and `NRSC5_EVENT_PACKET` events, which do include it). I think I'll add that into the API so that it's easier to filter LOT files, without having to cross-reference to the SIG table.

To solve your immediate problem of excessive SSD writes, you could perhaps ask nrsc5 to write to a tmpfs mount, and periodically copy out the files you're interested in.

argilo:
[...]

PR #407, opened 2025-05-07
Add data component MIME type to LOT event

argilo:
As a convenience, `NRSC5_EVENT_STREAM` and `NRSC5_EVENT_PACKET` events include the MIME type of the data component to which they belong. This makes it easier for applications to filter the data, without having to cross-reference the SIG table.

While investigating #295, I noticed that `NRSC5_EVENT_LOT` events lack this information. Here I've added a `component_mime` field to the event, which should make it easier for applications to filter LOT files.

The C and Python CLI applications now print this field:

```
[...]
19:35:48 SIG Service: type=data number=22 name=TTN STM
19:35:48   Data component: id=0 port=0810 service_data_type=66 type=3 mime=FF8422D7
19:35:48   Data component: id=2 port=0820 service_data_type=66 type=3 mime=EF042E96
[...]
19:35:48 LOT file: port=0820 lot=1633 name=DWRI_035iuu_rev04_0661.txt size=439 mime=BB492AAC component_mime=EF042E96 expiry=2024-06-12T15:42:00Z
[...]
19:35:49 LOT file: port=0820 lot=1624 name=DWRO_035iuu_rev04_20240612_1507_0658.png size=1526 mime=4F328CA0 component_mime=EF042E96 expiry=2024-06-12T15:42:00Z
[...]
```

```
[...]
9:37:56 SIG Service: type=DATA number=22 name=TTN STM
19:37:56   Data component: id=0 port=0810 service_data_type=IMAGE_MAPS type=3 mime=TTN_STM_TRAFFIC
19:37:56   Data component: id=2 port=0820 service_data_type=IMAGE_MAPS type=3 mime=TTN_STM_WEATHER
[...]
19:37:57 LOT file: port=0820 lot=1633 name=DWRI_035iuu_rev04_0661.txt size=439 mime=TEXT component_mime=TTN_STM_WEATHER expiry=2024-06-12T15:42:00Z
[...]
19:37:57 LOT file: port=0820 lot=1624 name=DWRO_035iuu_rev04_20240612_1507_0658.png size=1526 mime=PNG component_mime=TTN_STM_WEATHER expiry=2024-06-12T15:42:00Z
[...]
```

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -393,24 +393,25 @@ struct nrsc5_event_t
             uint16_t port;
             uint16_t seq;
             unsigned int size;
-            uint32_t mime;
+            uint32_t mime;       /**< MIME type of the data component this stream belongs to, e.g. NRSC5_MIME_HERE_IMAGE */
             const uint8_t *data;
         } stream;
         struct {
             uint16_t port;
             uint16_t seq;
             unsigned int size;
-            uint32_t mime;
+            uint32_t mime;       /**< MIME type of the data component this packet belongs to, e.g. NRSC5_MIME_NAVTEQ */
             const uint8_t *data;
         } packet;
         struct {
             uint16_t port;
             unsigned int lot;
             unsigned int size;
-            uint32_t mime;
+            uint32_t mime;           /**< MIME type of this file, e.g. NRSC5_MIME_PNG */
             const char *name;
             const uint8_t *data;
             struct tm *expiry_utc;
+            uint32_t component_mime; /**< MIME type of the data component this file belongs to, e.g. NRSC5_MIME_TTN_STM_WEATHER */
         } lot;
         struct {
             unsigned int program;       /**< program number 0, 1, ..., 7 */
--- a/src/main.c
+++ b/src/main.c
@@ -379,7 +379,8 @@ static void callback(const nrsc5_event_t *evt, void *opaque)
             dump_aas_file(st, evt);
         char time_str[64];
         strftime(time_str, sizeof(time_str), "%Y-%m-%dT%H:%M:%SZ", evt->lot.expiry_utc);
-        log_info("LOT file: port=%04X lot=%d name=%s size=%d mime=%08X expiry=%s", evt->lot.port, evt->lot.lot, evt->lot.name, evt->lot.size, evt->lot.mime, time_str);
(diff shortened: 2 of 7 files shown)


TheDaChicken:
At this point, you should also add component_channel or component_program. That way, I can cross reference the program id to the specific lot without needing to store it myself.

argilo:
I thought about that too. It's a bit more complicated since we'd need to grab the port number from the sibling audio component, but only if the SIG service is type=AUDIO. Maybe best to leave it to a separate PR.

TheDaChicken:
Wouldn't it be better to create a new field that holds all the component information in a struct than to add a single field for one piece of information? 

Adding a single field for components can result in being forced to add a ton of fields to the lot event in the future.

argilo:
Maybe? I suppose we could return a `nrsc5_sig_service_t *` and `nrsc5_sig_component_t *` and then the application could grab whatever it needs from there.

TheDaChicken:
Reusing a struct sounds like a good idea, in my opinion. 

In the future, the lot event could have these many fields instead of just two structs with all the information:

compont_mine
compont_id
compont_channel
compont_program

Etc

argilo:
OK, I think I'll leave this PR open for the time being and experiment with exposing `nrsc5_sig_service_t *` and `nrsc5_sig_component_t *` before we commit to an API addition.

I expect there's some value in making SIG data easily accessible, because both of the GUIs I looked at are relegated to linear searches through cached SIG data:

https://github.com/cmnybo/nrsc5-gui/blob/1113d5a16f3dba834c6ff378d35880433fbd752f/nrsc5_gui.py#L757-L771

https://github.com/TheDaChicken/nrsc5_gui/blob/607628c8a2ca7a904ff71ad87731f6d5de3769c4/src/HybridRadio.cpp#L529-L534

duracell80:
Appreciate the work on this! An API method or parameter flag is such a help in filtering.

Having ffmpeg compile a radar loop in conjunction with a time lapse USB wide angle camera in the window has been pretty fascinating.

For weather applications because the file size of a transparent PNG can indicate how much rain or snow is in the area, it's a pretty interesting thought to base a rain prediction calculation based on the file size alone. Maybe even CV2 could analyze the colors of the radar and judge if it's snowing or raining. Two interesting data points for a local weather station.

On the traffic side a data scientist could compile the location of the most frequent interstate jams by timestamp and hook up to public traffic RTSP streams.

argilo:
Closing in favour of #414.

PR #413, opened 2025-05-29
Increase the SIG component limit

argilo:
We're currently coming very close to hitting `MAX_SIG_COMPONENTS`; I've seen as many as 7. It's common to see as many as four data services, and stations can in theory have up to 8 audio services. To allow for some future expansion in the number of data services, I've increased the limit to 16.


argilo:
Oops, I meant to increase the service limit here, not the component limit. Fixed in 578bdb0ba8cc0d2747b74303cab0cdce9db6a3be.

PR #414, opened 2025-05-29
Include pointers to SIG service & component in packet/stream/LOT events

argilo:
Replaces #407.

The libnrsc5 API provides information about available data services in a single `NRSC5_EVENT_SIG` event. Later, data arrives in `NRSC5_EVENT_LOT`, `NRSC5_EVENT_STREAM` and `NRSC5_EVENT_PACKET` events.

Applications may want to know, for instance, whether a LOT file is a station logo or album art, and which audio stream it belongs to. The only way to achieve this is to store the information from the `NRSC5_EVENT_SIG` event, and later look it up by port number when an `NRSC5_EVENT_LOT` event arrives. It would be more convenient if the `NRSC5_EVENT_LOT` event included information about the service to which it belongs.

Here I've added `service` and `component` fields to `NRSC5_EVENT_LOT`, `NRSC5_EVENT_STREAM` and `NRSC5_EVENT_PACKET` events, thereby allowing easy lookup. For instance, in C:

```c
    if (evt->lot.component->data.mime == NRSC5_MIME_STATION_LOGO)
        log_info("Station logo for program %d", evt->lot.service->components->audio.port);
    if (evt->lot.component->data.mime == NRSC5_MIME_PRIMARY_IMAGE)
        log_info("Album art for program %d", evt->lot.service->components->audio.port);
```

Or in Python:
```python
    if evt.component.data.mime == nrsc5.MIMEType.STATION_LOGO:
        logging.info("Station logo for program %d", evt.service.components[0].audio.port)
    if evt.component.data.mime == nrsc5.MIMEType.PRIMARY_IMAGE:
        logging.info("Album art for program %d", evt.service.components[0].audio.port)
```

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -408,15 +408,19 @@ struct nrsc5_event_t
             uint16_t port;
             uint16_t seq;
             unsigned int size;
-            uint32_t mime;
+            uint32_t mime;  /**< DEPRECATED: Use `component->data.mime` instead */
             const uint8_t *data;
+            nrsc5_sig_service_t *service;
+            nrsc5_sig_component_t *component;
         } stream;
         struct {
             uint16_t port;
             uint16_t seq;
             unsigned int size;
-            uint32_t mime;
+            uint32_t mime;  /**< DEPRECATED: Use `component->data.mime` instead */
             const uint8_t *data;
+            nrsc5_sig_service_t *service;
+            nrsc5_sig_component_t *component;
         } packet;
         struct {
             uint16_t port;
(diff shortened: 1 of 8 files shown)


argilo:
@TheDaChicken noticed that it was necessary to access an audio service's audio component as `evt.service.components[0]`, which implicitly assumes that the audio component comes first in the SIG table. This is true of all stations I have seen, but we probably shouldn't count on it always being true.

To avoid the need to iterate over components to find the audio component, I've added an `audio_component` pointer to the `nrsc5_sig_service_t` structure, which points directly to the audio component, wherever it may be in the linked list. The audio component can then be accessed using `evt.service.audio_component`.

PR #415, opened 2025-05-30
Decode HERE Images

argilo:
Fixes #401.

Most of the details of the HERE Images stream were worked out in #308, so I've now added a decoder for it. Images are reported in `NRSC5_EVENT_HERE_IMAGE` API events, and logged by the C and Python command-line applications:

```
15:56:09 HERE Image: type=TRAFFIC, seq=2, n1=1, n2=9, time=2024-10-23T13:02:12Z, lat1=43.02372, lon1=-83.81881, lat2=42.82260, lon2=-83.54415, name=trafficMap_0_0_rdhs.png, size=3047
15:56:09 HERE Image: type=TRAFFIC, seq=2, n1=2, n2=9, time=2024-10-23T13:02:12Z, lat1=43.02372, lon1=-84.09347, lat2=42.82260, lon2=-83.26950, name=trafficMap_0_1_rdhs.png, size=4267
15:56:09 HERE Image: type=TRAFFIC, seq=2, n1=3, n2=9, time=2024-10-23T13:02:12Z, lat1=43.02372, lon1=-84.36813, lat2=42.82260, lon2=-82.99484, name=trafficMap_0_2_rdhs.png, size=5033
15:56:10 HERE Image: type=TRAFFIC, seq=2, n1=4, n2=9, time=2024-10-23T13:02:12Z, lat1=43.22419, lon1=-83.81881, lat2=42.62082, lon2=-83.54415, name=trafficMap_1_0_rdhs.png, size=4785
15:56:10 HERE Image: type=TRAFFIC, seq=2, n1=5, n2=9, time=2024-10-23T13:02:12Z, lat1=43.22419, lon1=-84.09347, lat2=42.62082, lon2=-83.26950, name=trafficMap_1_1_rdhs.png, size=6855
15:56:11 HERE Image: type=TRAFFIC, seq=2, n1=6, n2=9, time=2024-10-23T13:02:12Z, lat1=43.22419, lon1=-84.36813, lat2=42.62082, lon2=-82.99484, name=trafficMap_1_2_rdhs.png, size=5065
15:56:11 HERE Image: type=TRAFFIC, seq=2, n1=7, n2=9, time=2024-10-23T13:02:12Z, lat1=43.42400, lon1=-83.81881, lat2=42.41838, lon2=-83.54415, name=trafficMap_2_0_rdhs.png, size=3075
[...]

PR #419, opened 2025-05-31
Use 64-bit Unix timestamps for LOT file expiry & HERE file time

argilo:
In the libnrsc5 API, the LOT file expiry time is currently conveyed in a `struct tm`. This has turned out to be unwieldy, and mismatches the HERE Images API (added in #415) which uses a Unix timestamp to convey file time.

I propose to add a new `expiry_timestamp` field, and deprecate the old `expiry_utc` field. This also allows the code in the Python API wrapper to be simplified.

In addition, I think it would make sense to use `int64_t` to convey Unix timestamps in the API, since that's the underlying type of modern `time_t`.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -438,9 +438,10 @@ struct nrsc5_event_t
             uint32_t mime;
             const char *name;
             const uint8_t *data;
-            struct tm *expiry_utc;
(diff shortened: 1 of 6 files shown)


argilo:
Times beyond 2038 don't work on 32-bit Windows. This will come up soon, because file expiry times are often set 10 years in the future. Let me see what we can do about that.

argilo:
The updated code now works correctly on 32-bit Windows.

Issue #372, opened 2025-03-07
Expose debug information via the API.


pclov3r:
From https://github.com/theori-io/nrsc5/issues/254 it would be useful to expose debug information via the API.

Right now gain values that are set would be useful.

argilo:
With #490 merged, only two of the debug messages I listed above are left:

https://github.com/theori-io/nrsc5/blob/bc4901b8a9d27cd633f608e35e085d9801334b79/src/output.c#L369-L370

https://github.com/theori-io/nrsc5/blob/bc4901b8a9d27cd633f608e35e085d9801334b79/src/output.c#L697

argilo:
With #524 merged, the following are the most frequently seen debug messages across my library:

---

https://github.com/theori-io/nrsc5/blob/216b5cc11887ed8b663046b34d59bec032217d05/src/output.c#L762

This error occurs 162 times across a handful of recordings. The affected stations are continuing to broadcast station logos for subchannels that no longer exist. Since there's no way to know what the data stream means without a corresponding entry in the SIG table, there's little use in reporting anything through the API. We could perhaps quiet the logging by keeping track of which ports are in this state and only logging the error once per port.

---

https://github.com/theori-io/nrsc5/blob/216b5cc11887ed8b663046b34d59bec032217d05/src/output.c#L137

This error occurs 26 times, presumably when a corrupted audio packet makes it past the 8-bit CRC check by chance (which will occur with probability 1/256 for corrupt packets). In theory, it could also occur due to a bug in the patched FAAD2 library. It's possible these errors should be reported through the API in the same way as CRC errors.

---

https://github.com/theori-io/nrsc5/blob/216b5cc11887ed8b663046b34d59bec032217d05/src/frame.c#L408

This error occurs 7 times. I would guess that it occurs in low-SNR recodings where bit errors are frequent, but I haven't dug into it yet. The CCC (Configuration Control Channel) defines how the fixed bearer channel is subdivided into the data subchannels which carry AAS data. This is low-level information which may not be of any use to API clients.

argilo:
Decode errors are now reported in the API: #533

The library debug message remains, since it provides an informational string from the HDC decoder that may be of use to nrsc5 developers.

PR #428, opened 2025-06-03
Report frequency offset and service mode through the API

argilo:
As noted in #372, there is useful information in library debug messages which would be better conveyed through the API. Here I've moved over a couple of debug messages related to synchronization:

* The "CFO" message gave a rough estimate of the frequency offset. I replaced this with a more accurate calculation and put it into the `evt->sync.freq_offset` field.
* The Primary Service Mode Indicator (FM) / Service Mode Indicator (AM) are moved into the `evt->sync.psmi` field.

In the C and Python CLI applications, I logged these fields separately from the "Synchronized" log message so as not to break NRSC5-DUI's log parsing: https://github.com/markjfine/nrsc5-dui/blob/2d4908a8064bdcdf153303e36bb629de9bc677eb/nrsc5-dui.py#L305 (It really should switch to using the API!)

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -356,7 +356,7 @@ struct nrsc5_event_t
  * The member `event` determines which sort of event occurred:
  * - `NRSC5_EVENT_LOST_DEVICE` : RTL-SDR device was disconnected
  * - `NRSC5_EVENT_IQ` : IQ data, see the `iq` union member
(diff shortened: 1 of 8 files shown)


argilo:
The new log output looks like so:
```
14:16:01 Synchronized
14:16:01 Frequency offset: 293 Hz
14:16:01 Primary service mode: 1
```

PR #434, opened 2025-06-04
Move log.c to the "app" target

argilo:
Currently the C CLI app links statically to libnrsc5. Switching to dynamic linking causes the build to fail, because log.c is part of the libnrsc5 shared library, but not exported. I expect it was an oversight to put log.c in libnrsc5, because it's only the C CLI app that uses it.

To solve the problem, I've moved log.c into the source list for the "app" target.


argilo:
libnrsc5 does have some logging of its own, but it uses macros instead of log.c:

https://github.com/theori-io/nrsc5/blob/2fe91f323e3fa4a331820f21d4c0e9b31a3eeefc/src/defines.h#L74-L81

PR #435, opened 2025-06-05
Use hdrlen to check for LOT metadata header

argilo:
LOT fragments can either have a short header (with sequencing information only) or a long header (with sequencing information and file metadata). It just so happens that most stations use a long header for fragment 0 and a short header for all others, but this is not a requirement. The current parsing code assumes this is always the case, and uses the fragment sequence number to determine whether the header is short or long. We should instead determine that by checking the `hdrlen` field.


argilo:
I verified that all LOT files in my collection of recordings are decoded identically before & after this change.

PR #438, opened 2025-06-06
Don't report duplicate HERE images

argilo:
This is a follow-up to #415.

HERE images are sometimes sent multiple times. In other parts of the API we avoid reporting duplicate data, and I think we should do so here as well. This could help to avoid race conditions where HERE data is read while being overwritten by a duplicate copy.

FYI @markjfine


argilo:
To check for duplicates, I store 10 timestamps (one for weather, 9 for traffic tiles) and only report the image if the corresponding timestamp has changed.

PR #439, opened 2025-06-07
Refactor LOT LRU counter

argilo:
libnrsc5's LOT receiver can track up to 12 simultaneous file downloads per port. When a new LOT ID is seen, the least-recently-used LOT id is removed from the table of in-progress downloads to make space for a new one.

Currently, a static variable (`counter`) is incremented each time a LOT fragment arrives, and its value is stored in the corresponding LOT file struct. This value is used to determine which LOT ID is the least recently used. Unfortunately the static variable is shared across all `nrsc5_t` instances. Using a static variable without synchronization could lead to Undefined Behaviour in multi-threaded programs. To solve the problem, I propose to replace the static variable with a member variable in the `output_t` struct, which will be separate for each `nrsc5_t` instance.

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -145,6 +145,7 @@ static void aas_reset(output_t *st)
                     aas_free_lot(&component->data.lot_files[k]);
         }
     }
+    st->lot_lru_counter = 1;
 
     memset(st->services, 0, sizeof(st->services));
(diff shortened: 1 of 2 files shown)


argilo:
One small change: the counter must start at 1 because `timestamp == 0` is used to check for unused entries in the LOT table.

PR #441, opened 2025-06-20
Report AGC status in an API event

argilo:
Partially addresses #372.

Information about RTL-SDR automatic gain selection is only reported via library debug messages. Client applications are interested in this information, so it would make sense to expose it through an API event instead.

I've updated the C and Python CLI applications to display the best gain value at the "info" level, and the intermediate gain values at the "debug" level. (Use the `-l 1` argument to see them.)

```
19:43:56 Gain: 25.4 dB, Peak amplitude: -12.7 dBFS
19:43:56 Gain: 38.6 dB, Peak amplitude: 0.0 dBFS
19:43:56 Gain: 32.8 dB, Peak amplitude: -6.0 dBFS
19:43:56 Gain: 28.0 dB, Peak amplitude: -9.6 dBFS
19:43:56 Gain: 29.7 dB, Peak amplitude: -8.5 dBFS
19:43:56 Best gain: 29.7 dB, Peak amplitude: -8.5 dBFS
```

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -183,7 +183,8 @@ enum
     NRSC5_EVENT_EMERGENCY_ALERT,
     NRSC5_EVENT_HERE_IMAGE,
     NRSC5_EVENT_LOT_HEADER,
-    NRSC5_EVENT_LOT_FRAGMENT
+    NRSC5_EVENT_LOT_FRAGMENT,
+    NRSC5_EVENT_AGC
 };
 
 enum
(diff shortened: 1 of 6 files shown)


argilo:
@markjfine Note that NRSC5-DUI would need to have its regular expression updated to allow the timestamp in front of the "Best gain" message. If that's the only library debug message that NRSC5-DUI depends on, then the instruction to build nrsc5 with `-DLIBRARY_DEBUG_LEVEL=1` could be removed from the README.

markjfine:
Cheers.

PR #442, opened 2025-07-07
make CLI executable optional, allow use of system FAAD by indicating it's patched already

eean:
The purpose of making CLI executable optional is to avoid the libao dependency.

The INSTALLED_FAAD_IS_PATCHED option makes it possible to build nrsc5 without the Internet access required by ExternalProject. 

Together I'm using these commits to build my WIP [nrsc5 frontend in a Flatpak](https://invent.kde.org/ianmonroe/digital-terrestrial-radio/-/blob/master/.flatpak-manifest.json).


pclov3r:
To add to this there was an idea to upstream the FAAD2 patch but it will probably need to be cleaned up if and when that happens.

PR #443, opened 2025-07-17
Add 16-bit input to the C CLI. 

pclov3r:
This brings the C CLI in line with the python CLI using nrsc5_pipe_samples_cu8 and nrsc5_pipe_samples_cs16 for reading a i/q file. 

Also adds support for 16-bit input via a new --iq-input-format option to select 8 or 16-bit

This was done screwing around with a AI model so it needs to be reviewed *carefully*. Especially around threading it it's induced a bug because changes were made there. I decided to make a PR just to throw it out there as an idea. 

It seems to function correctly here but some more testing and review is required. 

IIRC the goal was try and bring the C and python CLI to be a mirror. 

Thanks!

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -25,6 +25,7 @@
 #include <limits.h>
 #include <stdio.h>
 #include <string.h>
+#include <errno.h> // For providing detailed error messages when file I/O fails.
 
 #ifdef __MINGW32__
 #include <conio.h>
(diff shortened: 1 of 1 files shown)


pclov3r:
Appears it also added more verbose output to explain the command options. 

As said this is just a draft and idea not sure if we really want to add that or not.

pclov3r:
I went ahead and just updated the command line options to reflect what's actually in the readme. Just to keep it consistent if that is added or not.

PR #444, opened 2025-07-21
Add IQ files to the first paragraph in the readme. 

pclov3r:
A simple change to the readme. Added IQ files to the first paragraph in the readme so it's a bit more clear that this isn't only just for the rtl-sdr. 

Maybe we need to include the required sample rate too?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # nrsc5
 
-This program receives NRSC-5 digital radio stations using an RTL-SDR dongle. It offers a command-line interface as well as an API upon which other applications can be built. Before using it, you'll first need to compile the program using the build instructions below.
+This program receives NRSC-5 digital radio stations using an RTL-SDR dongle or from IQ files. It offers a command-line interface as well as an API upon which other applications can be built. Before using it, you'll first need to compile the program using the build instructions below.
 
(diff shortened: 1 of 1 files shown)


pclov3r:
To add to this the About description should be changed to to add this but you can't do a PR for that :)

README.md
```diff
@@ -1,6 +1,6 @@
 # nrsc5
 
-This program receives NRSC-5 digital radio stations using an RTL-SDR dongle. It offers a command-line interface as well as an API upon which other applications can be built. Before using it, you'll first need to compile the program using the build instructions below.
+This program receives NRSC-5 digital radio stations using an RTL-SDR dongle or from IQ files. It offers a command-line interface as well as an API upon which other applications can be built. Before using it, you'll first need to compile the program using the build instructions below.
```
argilo, line 3: How does this version sound? I hope it's a bit clearer.
```suggestion
This program receives NRSC-5 digital radio stations using an RTL-SDR dongle, or by reading from I/Q files. It offers a command-line interface as well as an API upon which other applications can be built. Before using it, you'll first need to compile the program using the build instructions below.
```

pclov3r:
I think it's better and the changes are now applied in this PR :)

Issue #454, opened 2025-10-16
HD EAS Segment fault


TheDaChicken:
Today, all of Washington and a bunch of other states in the west coast had their required monthly tests today at 10:16AM.

I recorded two stations here KQMV-FM and KPNW-FM. I tried to record both at the same time. I ended up getting a segment fault on nrsc5. I had to switch recording from nrsc5 to SDR# so I would get a few minutes of recording. There is some error in decoding the control data. 

IQ recordings here:
https://www.dropbox.com/scl/fi/pm72xmkrszqsrepi58i0b/KQMV-FM-EAS-10_16_25-10_20.cu8?rlkey=r499wle16yuz5l02kexgkntyi&st=a03v5onz&dl=1
https://www.dropbox.com/scl/fi/m1cl7e5v9vtdncful522h/KNPW-FM-EAS-10_16_25-10_44AM.cu8?rlkey=1fsgyizsongw8ftzsa0moulgt&st=yvw2lrs8&dl=1

GDB says this:
```gdb
Thread 7 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 25372.0x9078]
0x00007ff76da41afd in callback (evt=0x35df100, opaque=<optimized out>)
    at C:/msys64/home/Aidan/nrsc5/src/main.c:497
warning: 497    C:/msys64/home/Aidan/nrsc5/src/main.c: No such file or directory
Quit)
(gdb) where
#0  0x00007ff76da41afd in callback (evt=0x35df100, opaque=<optimized out>)
    at C:/msys64/home/Aidan/nrsc5/src/main.c:497
#1  0x00007ff76da44d8c in nrsc5_report (st=<optimized out>, evt=0x35deeac)
    at C:/msys64/home/Aidan/nrsc5/src/nrsc5.c:655
#2  nrsc5_report_emergency_alert (st=<optimized out>,
    message=message@entry=0x181cc70 "The National Weather Service has issued a Required Monthly Test for Mason, WA, Kitsap, WA, Snohomish, WA, Pierce, WA, and King, WA beginning at 10:16 am and ending at 1:16 pm (KQMV)",
    control_data=0x7efefefefefefeff <error: Cannot access memory at address 0x7efefefefefefeff>,
    control_data@entry=0x209a5e0 "", control_data_length=1513899084, control_data_length@entry=17,
    category1=30, category2=30, location_format=0, num_locations=5, locations=locations@entry=0x0)
[...]

PR #455, opened 2025-10-17
Fix off-by-one error in EAS length checks

argilo:
Partially addresses #454.

The root cause of the crash @TheDaChicken encountered in #454 is that the bounds checks in the EAS location decoder are off by one. If the location data occupies all control data bits, then the final bounds check incorrectly fails, preventing the location data from being reported.

The crash revealed a second issue: if the control data were in fact invalid and a bounds check legitimately failed, then `decode_control_data` exits early and does not return a pointer to an array of locations. But it does return the decoded `num_locations` value, and so the API client fails when it attempts to read the non-existent array. I'll think about how best to solve this and open a follow-up PR.

### DIFF
--- a/src/pids.c
+++ b/src/pids.c
@@ -207,7 +207,7 @@ static void decode_control_data(const char *control_data, int len, int *category
 
     for (int i = 0; i < *num_locations; i++)
     {
-        if (off + 1 >= len*8)
+        if (off + 1 > len*8)
         {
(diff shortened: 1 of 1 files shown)


argilo:
This fixes the crash with the recordings in #454, and doesn't appear to change the behaviour of the receiver otherwise (when tested against my library of recordings).

PR #470, opened 2025-12-22
Provide CRC errors in API 

TheDaChicken:
This is my implementation of providing CRC errors in the API.

I moved the logic for partial packets into the elastic buffer. This is to hopefully reduce code complexity. Elastic buffer is basically another packet store.

I also made a packet_ref_t to hopefully reduce argument calls when pushing packets into the elastic buffer.

### DIFF
--- a/support/cli.py
+++ b/support/cli.py
@@ -26,6 +26,7 @@ class NRSC5CLI:
         self.wav_output = None
         self.raw_output = None
         self.hdc_output = None
+        self.audio_packets_valid = 0
         self.audio_packets = 0
         self.audio_bytes = 0
         self.audio_errors = 0
@@ -229,14 +230,18 @@ class NRSC5CLI:
                 self.audio_bytes += len(evt.data)
                 if evt.flags & nrsc5.PacketFlags.CRC_ERROR:
                     self.audio_errors += 1
+                else:
+                    self.audio_packets_valid += 1
 
-                if self.audio_packets >= 32:
+                if self.audio_packets_valid >= 32:
                     logging.info("Audio bit rate: %.1f kbps", self.audio_bytes * 8 * nrsc5.SAMPLE_RATE_AUDIO
-                                 / nrsc5.AUDIO_FRAME_SAMPLES / self.audio_packets / 1000)
+                                 / nrsc5.AUDIO_FRAME_SAMPLES / self.audio_packets_valid / 1000)
+                    self.audio_packets_valid = 0
+                    self.audio_bytes = 0
+                if self.audio_packets >= 32:
                     if self.audio_errors > 0:
                         logging.warning("Audio packet CRC mismatches: %d", self.audio_errors)
                     self.audio_packets = 0
-                    self.audio_bytes = 0
                     self.audio_errors = 0
         elif evt_type == nrsc5.EventType.AUDIO:
             if evt.program == self.args.program:
--- a/src/main.c
+++ b/src/main.c
@@ -72,6 +72,7 @@ typedef struct {
 
     unsigned int program;
     unsigned int audio_ready;
+    unsigned int audio_packets_valid;
     unsigned int audio_packets;
     unsigned int audio_bytes;
     unsigned int audio_errors;
@@ -347,13 +348,19 @@ static void callback(const nrsc5_event_t *evt, void *opaque)
             st->audio_bytes += evt->hdc.count * sizeof(evt->hdc.data[0]);
             if (evt->hdc.flags & NRSC5_PKT_FLAGS_CRC_ERROR)
                 st->audio_errors++;
+            else
+                st->audio_packets_valid++;
 
-            if (st->audio_packets >= 32) {
-                log_info("Audio bit rate: %.1f kbps", (float)st->audio_bytes * 8 * NRSC5_SAMPLE_RATE_AUDIO / NRSC5_AUDIO_FRAME_SAMPLES / st->audio_packets / 1000);
+            if (st->audio_packets_valid >= 32) {
+                log_info("Audio bit rate: %.1f kbps", (float)st->audio_bytes * 8 * NRSC5_SAMPLE_RATE_AUDIO / NRSC5_AUDIO_FRAME_SAMPLES / st->audio_packets_valid / 1000);
+                st->audio_packets_valid = 0;
+                st->audio_bytes = 0;
+            }
+            if (st->audio_packets >= 32)
+            {
                 if (st->audio_errors > 0)
                     log_warn("Audio packet CRC mismatches: %d", st->audio_errors);
                 st->audio_packets = 0;
-                st->audio_bytes = 0;
                 st->audio_errors = 0;
             }
         }


argilo:
Some thoughts on logging:

1. Since the CLI apps only play back one program at a time, it would probably make sense to log only for HDC packets pertaining to the selected program. If we do that, then the program number will no longer need to be part of the log messages.
2. The C & Python CLIs should use the same log level. (At the moment, C logs at the warning level, and Python at the error level).
3. CRC errors are normal at low SNR, so they should probably be reported at most at the warning level.
4. CRC errors can be frequent at low SNR, so it might be worth aggregating them and logging a count. The CLI apps report bit rate once every 32 packets, so it might make sense to do the same for CRC errors.
5. It's entirely normal and expected to see a half-packet when synchronization is achieved. I expect the same would be true when synchronization is lost, and half-packets shouldn't occur in other circumstances. Since they are normal and expected, they probably should be logged at the debug level. In fact, I'm not convinced that they should be reported through the API at all, but I'll have to think about it some more.

argilo:
> I can always remove it. I only included it to bring back the original behavior since "ignoring partial pdu" used to be logged. I cannot think of any other reason to include it.

I suspect the old "ignoring partial pdu" message (removed in #427) was just there to help with debugging during the initial development of nrsc5. I didn't ever need it for anything myself, even while debugging.

After thinking about it some more, I don't think it's worth including the `HALF_BACK` and `HALF_FRONT` flags in the API. I can't think of any use that an API client could have for knowing the details of how a particular audio packet was packed into audio transport PDUs.

argilo:
The latter. I don't see a need to report anything other than CRC errors.

TheDaChicken:
Okay, I removed reporting half packets. This also now doesn't save packets with crc errors since there is no point to keep them in elastic-buffer. 

I also squashed commits to make looking at the changes a lot easier.

argilo:
Thanks. I'll do a bit more testing and a final code review, but it may not be until Monday.

src/output.c
```diff
@@ -34,18 +34,67 @@ void output_align(output_t *st, unsigned int program, unsigned int stream_id, un
...
+        {
+            pkt->size = 0;
+        }
+    }
+}
+
+static void pkt_reset(packet_t* pkt)
+{
+    pkt->size = 0;
+    pkt->flags = 0;
```
argilo:
```suggestion
    pkt->flags = PACKET_FLAG_NONE;
```

support/cli.py
```diff
@@ -226,11 +227,17 @@ def callback(self, evt_type, evt):
...
                 self.audio_packets += 1
                 self.audio_bytes += len(evt.data)
+                if evt.flags & nrsc5.PacketFlags.CRC_ERROR:
+                    self.audio_errors += 1
+
                 if self.audio_packets >= 32:
                     logging.info("Audio bit rate: %.1f kbps", self.audio_bytes * 8 * nrsc5.SAMPLE_RATE_AUDIO
                                  / nrsc5.AUDIO_FRAME_SAMPLES / self.audio_packets / 1000)
+                    if self.audio_errors > 0:
+                        logging.warning("CRC mismatch: %d/%d", self.audio_errors, self.audio_packets)
```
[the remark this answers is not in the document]
argilo: The "/32" denominator could be a bit misleading since it's only printed when the numerator is greater than zero. Maybe we don't need to include it?

src/nrsc5.c
```diff
@@ -702,14 +702,24 @@ void nrsc5_report_lost_sync(nrsc5_t *st)
...
+    evt.hdc.count = 0;
+    evt.hdc.flags = 0;
+
+    if (pkt->shape == PACKET_FULL)
+    {
+        evt.hdc.data = pkt->data;
+        evt.hdc.count = pkt->size;
+    }
+    if (pkt->flags & PACKET_FLAG_CRC_ERROR)
+        evt.hdc.flags |= NRSC5_PKT_FLAGS_CRC_ERROR;
```
argilo: I assume the separation here is so that we could add flags in the future that are only used for internal bookkeeping inside the library?

src/output.c
```diff
@@ -62,16 +111,18 @@ void output_advance(output_t *st)
...
 #ifdef USE_FAAD2
             int produced_audio = 0;
 #endif
 
-            if (len > 0)
+            if (is_complete_pkt(pkt))
             {
-                nrsc5_report_hdc(st->radio, program, pkt, len);
+                nrsc5_report_hdc(st->radio, program, pkt);
+            }
```
argilo: Why do we skip reporting incomplete packets here?

argilo: Wait, maybe I misunderstood something. I noticed a lot more HDC packets getting reported for my low-SNR samples, and assumed we're now intentionally reporting empty HDC packets. But maybe that's not intentional?

argilo: Never mind. After thinking some more I realized that the "extra" HDC packets must be the ones with bad CRCs, which were not previously reported.

argilo: It may be worth considering whether we should report "empty" HDC packets in the cases where we return silence in the audio stream.

TheDaChicken: Oh, I think understand the confusion. I wanted empty HDC packets with CRC flag set to be reported. Is that not what is happening? 

I am not sure if you have another idea. I thought it would make sense to report HDC packets or empty HDC packets with CRC flag only if they were complete packets.

support/cli.py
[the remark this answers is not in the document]
TheDaChicken: I included /32 because I wanted to provide context on how serve the error count is relative to the amount of audio decoded. I don't think printing out the amount of audio errors provides that context.

src/nrsc5.c
TheDaChicken: Yes. I do this by habit. I could switch to using NRSC5_PKT_FLAGS_CRC_ERROR internally within the library if you want me too.

src/output.c
```diff
@@ -34,18 +34,67 @@ void output_align(output_t *st, unsigned int program, unsigned int stream_id, un
...
+        log_warn("Packet %d already exists in elastic buffer for program %d, stream %d. Overwriting.", ref->seq, ref->program, ref->stream_id);
+
+    if (ref->shape == PACKET_HALF_BACK && pkt->shape == PACKET_HALF_FRONT)
+    {
+        pkt->flags |= ref->flags;
+        pkt->shape = PACKET_FULL;
 
-    memcpy(elastic->packets[seq].data, pkt, len);
-    elastic->packets[seq].size = len;
+        if (is_crc_ok(pkt))
```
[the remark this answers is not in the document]
TheDaChicken: I can understand this isn't clear. I can rewrite this code or if you have better way to rewrite this code.

Before this line, I add the ref flags to pkt flags. If the pkt is already corrupted, I don't want to save more corrupted data to the packet. If the ref packet is corrupted, mark it as corrupted with a data size of 0.

support/cli.py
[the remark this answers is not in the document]
argilo: The problem I see is that the context could be misleading. If the actual CRC error rate was, say, 1/100, the user would see 1/32 from time to time, and occasionally 2/32. From that, one might conclude that the error rate was somewhere between 1/32 and 2/32. Printing out 0/32 would solve the problem, but it would unnecessarily clutter the logs.

src/output.c
argilo: I do think that reporting CRC errors in HDC events makes sense, and it's the same approach I was planning to take before you took care of it.

I think ideally the `NRSC5_EVENT_HDC` and `NRSC5_EVENT_AUDIO` streams should line up packet-for-packet so that the audio stream could be exactly recovered from the HDC stream if desired. But I don't think we're quite there yet, since there are circumstances where we skip reporting an HDC event.

We could leave that to a future PR if you like.

TheDaChicken: Ohhhhh, okay. I understand now. I think that should be in a separate pr. I can work on that after this. It'll be able to discuss with you my thoughts on that.

support/cli.py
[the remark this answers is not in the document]
argilo: Maybe packet error rate could be displayed alongside the bit error rate in the CLI?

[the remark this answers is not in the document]
TheDaChicken: Well, an average would be counted based on the count of total packets received. That may cause issues when changing channels unless I count and only print based on the current program?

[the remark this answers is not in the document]
argilo: > I could add the error count to the bitrate message. That would mean the message is bigger and the error counts on the info log level

Yeah, reporting in the BER message has some downsides, and it's a bit awkward to report stuff that came from an HDC event while logging a BER event.

Maybe it's best to just report the CRC error count, and not bother trying to express it as a rate.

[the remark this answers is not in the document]
TheDaChicken: Funny you say that because that's what average BER is biased then since it does the same thing. 

Alright. I will remove "/32" from the log message.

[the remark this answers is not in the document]
TheDaChicken: > > Funny you say that because that's what average BER is biased then since it does the same thing.
> 
> It's not, because it's printed all the time, not only when bit errors occur.

Ohhh!!!! Okay, that makes sense.

TheDaChicken:
Okay I updated changes based on your feedback.

Apoglizes if you saw my comments posted twice. The Github mobile app is sending my comments twice for some reason.

argilo:
One last issue I just noticed: the presence of zero-length HDC packets pushes down the calculated audio bit rate:
```
20:09:18 Audio bit rate: 0.0 kbps
20:09:18 Audio packet CRC mismatches: 32
--
20:09:18 Audio bit rate: 0.0 kbps
20:09:18 Audio packet CRC mismatches: 32
--
20:09:18 Audio bit rate: 2.4 kbps
20:09:18 Audio packet CRC mismatches: 31
--
20:09:18 Audio bit rate: 1.7 kbps
20:09:18 Audio packet CRC mismatches: 31
--
20:09:19 Audio bit rate: 3.6 kbps
20:09:19 Audio packet CRC mismatches: 30
--
20:09:19 Audio bit rate: 0.0 kbps
20:09:19 Audio packet CRC mismatches: 32
```
I'll have to think about how best to handle that.

TheDaChicken:
> One last issue I just noticed: the presence of zero-length HDC packets pushes down the calculated audio bit rate:
> 
> ```
> 20:09:18 Audio bit rate: 0.0 kbps
> 20:09:18 Audio packet CRC mismatches: 32
> --
> 20:09:18 Audio bit rate: 0.0 kbps
> 20:09:18 Audio packet CRC mismatches: 32
> --
> 20:09:18 Audio bit rate: 2.4 kbps
> 20:09:18 Audio packet CRC mismatches: 31
> --
> 20:09:18 Audio bit rate: 1.7 kbps
> 20:09:18 Audio packet CRC mismatches: 31
> --
> 20:09:19 Audio bit rate: 3.6 kbps
> 20:09:19 Audio packet CRC mismatches: 30
> --
> 20:09:19 Audio bit rate: 0.0 kbps
> 20:09:19 Audio packet CRC mismatches: 32
> ```
> 
> I'll have to think about how best to handle that.

Oh crap, you are right. I think if you wanted to report NRSC5_EVENT_HDC and NRSC5_EVENT_AUDIO packet-for-packet so that the audio stream could be exactly recovered from the HDC stream if desire, I think this should be dealt with instead of creating a new event for crcs.

argilo:
Another behaviour change in the API clients is that `--dump-hdc` will write out an ADTS header & zero-length packet each time an invalid CRC occurs. Maybe that's a sensible thing to do though, if we're aiming for enough information to be saved that the audio stream can be fully recovered.

TheDaChicken:
Actually, no. Ignore what I am saying. This would be fine actually.

TheDaChicken:
Is something like this fine? To count valid packets and all audio_packets??

```python
        elif evt_type == nrsc5.EventType.HDC:
            if evt.program == self.args.program:
                if self.args.dump_hdc:
                    self.hdc_output.write(self.adts_header(len(evt.data)))
                    self.hdc_output.write(evt.data)

                self.audio_packets += 1
                self.audio_bytes += len(evt.data)
                if evt.flags & nrsc5.PacketFlags.CRC_ERROR:
                    self.audio_errors += 1
                else:
                    self.audio_packets_valid += 1

                if self.audio_packets_valid >= 32:
                    logging.info("Audio bit rate: %.1f kbps", self.audio_bytes * 8 * nrsc5.SAMPLE_RATE_AUDIO
                                 / nrsc5.AUDIO_FRAME_SAMPLES / self.audio_packets_valid / 1000)
                    self.audio_packets_valid = 0
                    self.audio_bytes = 0
                    self.audio_errors = 0
                if self.audio_packets >= 32:
                    if self.audio_errors > 0:
                        logging.warning("Audio packet CRC mismatches: %d", self.audio_errors)
                    self.audio_packets = 0
```

argilo:
Yeah, I think that would work. I don't think it's a big deal if CRC errors are reported at a different cadence as the audio bit rate.

argilo:
Looking better, I'm now seeing the expected output from both clients.

TheDaChicken:
I am so glad you do testing. Thanks for checking. I do this thing where I only test on either the python cli or the c client. 😄 I rebased and squashed two commits trying to fix it. Should work now.

TheDaChicken:
Thank you!

TheDaChicken:
WOW! That's a lot of logging!

PR #473, opened 2025-12-31
Find the path of faad2_hdc

TheDaChicken:
Whenever I wanted to make changes to HDC patch, I had to save it as a patch and let cmake recompile it.

`INSTALLED_FAAD_IS_PATCHED` option has made it easier. Currently, it assumes that the linker will find faad_hdc. I cannot provide a location if the linker cannot find it. 

You would normally provide a cmake find module file. Currently nrsc5's cmake relies on only pkgconfig. I think switching to cmake find module files would be a separate PR I think.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,14 +220,24 @@ endif ()
 if (USE_FAAD2)
     # use patched faad if installed
     if (INSTALLED_FAAD_IS_PATCHED)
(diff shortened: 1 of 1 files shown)


TheDaChicken:
@eean I hope this still would work for you since you originally made the `INSTALLED_FAAD_IS_PATCHED` option.

PR #475, opened 2026-01-03
Collect & Provide enhanced packets through the API

TheDaChicken:
libnrsc5 used to report HDC enhanced packets before elastic-buffer changes. The enhacned packets weren't aligned together because the elastic-buffer didn't exist. This is meant to align them via elastic buffer and collects the enhanced packets to bring them together with the HDC core stream. 

100% looking for feedback. I am not sure if enhanced packets should be part of bitrate calculation etc. I do print another log message for mismatches in for enhanced packets. 

This is where maybe an flag needs to be added to say the enhanced stream isn't being decoded? That way, clients can ignore this for bitrate?

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -414,6 +414,9 @@ struct nrsc5_event_t
             const uint8_t *data;
             size_t count;
             unsigned int flags; /** The specific status of the hdc packet. Example `NRSC5_PKT_FLAGS_CRC_ERROR` **/
+            const uint8_t *enh_data;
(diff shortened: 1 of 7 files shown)


TheDaChicken:
Another idea for this PR is to instead:

deprecate the old data field and switch to single array of structs that can point to data, size and flags. This could be helpful when more streams get added because it wouldn't require much changes. 

The downside is deprecating another field. The HD spec does reserve another two streams. In my opinion, the possibility of those two other streams to be used is small.

PR #477, opened 2026-01-05
In MA1 mode, delay the P3 logical channel to match P1

argilo:
In service mode MA1, a diversity delay of 3 frames is applied to the P1 logical channel (specifically targeting the BL and BU subframes on the transmit side). As a result, the P1 logical channel has an additional 3 frames of latency and the receiver must delay the ML and MU subframes before deinterleaving so that all subframes have a combined (TX + RX) delay of exactly 3 frames.

The P3 logical channel does not have a diversity delay. To keep the received data aligned with the P1 logical channel (so that the core & enhanced streams can be coherently combined), it is necessary to apply an additional 3-frame delay in the receiver. I've accomplished that here by applying a 3-frame delay to the EL and EU subframes that carry the P3 logical channel.

No change is required for service mode MA3, because its P3 logical channel already has a 3-frame delay applied to the EML and EMU subframes, complementing the transmitter's 3-frame delays to the EBL and EBU subframes.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -43,7 +43,7 @@ static int bit_map(unsigned char matrix[PARTITION_WIDTH_AM * BLKSZ * 8], int b, 
 static void interleaver_ma1(decode_t *st)
 {
     int b, k, p;
-    for (int n = 0; n < 18000; n++)
+    for (int n = 0; n < BL_LENGTH; n++)
     {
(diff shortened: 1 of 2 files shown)


argilo:
I also replaced the magic number 18000 in a bunch of places, because not all subframes have the same length.

Issue #461, opened 2025-11-14
Should libnrsc5's Input buffer be removed?


TheDaChicken:
I was looking at the code for possible areas of refactoring. 

I think it may be a good idea to remove the input buffer in libnrsc5. This way, the idea that libnrsc5 processes per block is more clearly shown and integrated by processing the user input block by block. Leftovers can still be stored in acquire. We can ask Acquire if we have a full block or not. This also reduces the stack memory 

The input buffer size also isn't documented. libnrsc5 could silently overflow without providing the caller any error. 

This may also allow easier improvements of the elastic buffer. It may be more clear how many samples is leftover when processing.

PR #481, opened 2026-03-29
Remove `input_t.buffer`

TheDaChicken:
Deals with #461 

This PR removes the huge buffer in input.c. To do this, this PR does:
- For nrsc5's 8 bits format, ~~input will now push into acquire one at a time~~ input will now will push blocks into acquire. 
- For nrsc5's 16 bits format, input will now push data directly into acquire. 

No unnecessary extra copying and input size constraints

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -38,22 +38,22 @@ static float decim_taps[] = {
     -0.00410953676328063
 };
 
-void input_push(input_t *st, const cint16_t* buf, const uint32_t length)
+void input_push(input_t *st, const cint16_t* buf, const uint32_t len)
 {
-    uint32_t consumed = 0;
+    unsigned int consumed = 0;
 
-    while (consumed < length)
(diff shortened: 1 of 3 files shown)


src/input.c
```diff
@@ -114,32 +84,46 @@ void input_push_cu8(input_t *st, const uint8_t *buf, uint32_t len)
...
-    input_push(st);
+    return avail;
 }
 
-void input_push_cs16(input_t *st, const int16_t *buf, uint32_t len)
+void input_push_cu8(input_t *st, const uint8_t *buf, const uint32_t len)
 {
-    assert(len % 2 == 0);
+    assert(len % 4 == 0);
+    cint16_t out[FFTCP_FM];
```
argilo: I expect we can halve the size of this buffer, since the decimation factor is either 2 (FM) or 32 (AM).

PR #484, opened 2026-03-31
Add AM system control values to the API and CLI apps

argilo:
For AM stations, libnrsc5 currently reports the Service Mode Indicator (SMI) in SYNC events. Alongside this value, the reference subcarrier also conveys four others:

* Power Level Indicator (PLI)
* High-Power PIDS Indicator (HPPI)
* Analog Audio Bandwidth Indicator (AABI)
* Reduced Digital Bandwidth Indicator (RDBI)

Descriptions of these values can be found in NRSC-5-E, Layer 1 AM, section 11.2.2 - 11.2.5.

I think these values are interesting enough to expose in the API, so I've done so here.


argilo:
gr-nrsc5 can transmit all the corresponding modes, so I used it to verify that the indicators are correctly decoded and displayed in both the C and Python CLI applications.

PR #485, opened 2026-04-01
Adds polling for stdin onto linux & windows & remove pthread_cancel etc.

### LINKED ISSUE #314: On Windows, NRSC5 should quit after reading an I/Q file

TheDaChicken:
To fix #314 (wow, all the back in 2023) and related to #394 

Instead of using read() which waits for input to be typed in stdin, we can use polling with a timeout to allow threads to properly release when st->closed is set. This PR uses linux/unix and windows specific APIs to do this. 

This PR also removes using `pthread_cancel.` This function isn't really recommended and isn't needed when polling.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -36,6 +36,7 @@
 #include <netdb.h>
 #include <sys/socket.h>
 #include <termios.h>
+#include <poll.h>
 #endif
 
 #include "bitwriter.h"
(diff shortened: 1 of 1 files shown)


argilo:
I tested this on Linux, macOS, and Windows. I saw no problems with exiting, whether playing back an I/Q file or receiving a live station.

It's nice that getting rid of `pthread_cancel` also removes the need for `pthread_cleanup_push`.

The <= 100ms polling delay when reaching the end of a recording seems acceptable. While it might be nice to avoid polling, I suspect the design would become more complicated. So I think the current design is fine.

PR #490, opened 2026-04-14
Exciter/Importer, Leap Offset & Local Time to API

TheDaChicken:
This PR is to move another five debug messages in library to the API.  #372 

This includes four new events:

- Importer Info
- Exciter Info
- Leap offset
- Local Time

One silly new parameter that has not been logged is importer_connected. This also combines all leap offsets into one debug print.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -185,7 +185,9 @@ enum
     NRSC5_EVENT_HERE_IMAGE,
     NRSC5_EVENT_LOT_HEADER,
     NRSC5_EVENT_LOT_FRAGMENT,
-    NRSC5_EVENT_AGC
+    NRSC5_EVENT_AGC,
+    NRSC5_EVENT_EXCITER_INFO,
+    NRSC5_EVENT_IMPORTER_INFO,
 };
 
 enum
@@ -391,6 +393,8 @@ struct nrsc5_event_t
  * - `NRSC5_EVENT_EMERGENCY_ALERT` : emergency alert, see `emergency_alert` member
  * - `NRSC5_EVENT_HERE_IMAGE` : HERE Images traffic/weather map, see `here_image` member
  * - `NRSC5_EVENT_AGC` : automatic gain control status, see `agc` member
+ * - `NRSC5_EVENT_EXCITER_INFO` : exciter data, see `device_info` member
+ * - `NRSC5_EVENT_IMPORTER_INFO` : importer data, see `device_info` member
  */
     unsigned int event;
     union
@@ -573,6 +577,14 @@ struct nrsc5_event_t
             float peak_dbfs;     /**< peak signal amplitude in dB, relative to full scale */
             int is_final;        /**< 1 if this is the final (best) gain value, otherwise 0 */
         } agc;
+        struct {
+            char id[2];                  /**< 2-character identifier of the device, e.g. "GG" or "L7" */
+            int core_version[4];         /**< Core Version number. */
+            int manufacturer_version[4]; /**< Manufacturer-assigned Version number. */
+            int core_status;             /**< Core status. Values such like 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
+            int manufacturer_status;     /**< Manufacturer status. Values such like 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
+            int importer_connected;      /**< 1 if an importer is connected, otherwise 0 for EXCITER_INFO event. Set to -1 for IMPORTER_INFO event. */
+        } device_info;
     };
 };
 /**
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -188,6 +188,8 @@ enum
     NRSC5_EVENT_AGC,
     NRSC5_EVENT_EXCITER_INFO,
     NRSC5_EVENT_IMPORTER_INFO,
+    NRSC5_EVENT_LEAP_OFFSET,
+    NRSC5_EVENT_LOCAL_TIME,
 };
 
 enum
@@ -395,6 +397,8 @@ struct nrsc5_event_t
  * - `NRSC5_EVENT_AGC` : automatic gain control status, see `agc` member
  * - `NRSC5_EVENT_EXCITER_INFO` : exciter data, see `device_info` member
  * - `NRSC5_EVENT_IMPORTER_INFO` : importer data, see `device_info` member
+ * - `NRSC5_EVENT_LEAP_OFFSET` : leap offset, see `leap_offset` member
+ * - `NRSC5_EVENT_LOCAL_TIME` : local time data, see `local_time` member
  */
     unsigned int event;
     union
@@ -585,6 +589,17 @@ struct nrsc5_event_t
             int manufacturer_status;     /**< Manufacturer status. Values such like 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
             int importer_connected;      /**< 1 if an importer is connected, otherwise 0 for EXCITER_INFO event. Set to -1 for IMPORTER_INFO event. */
         } device_info;
+        struct {
+            int pending_leap_offset;          /**< Future GPS-UTC offset in seconds. Meant to be broadcasted months before the leap seconds and a few hours afterward. */
+            int current_leap_offset;          /**< Current GPS-UTC offset in seconds. */
+            int alfn_pending_leap_adjustment; /**< Represents GPS time of a pending leap second adjustment in ALFN.*/
+        } leap_offset;
+        struct {
+            int utc_offset;    /**< Local Time Zone UTC Offset in MINUTES. */
+            int dst_regional;  /**< 1 if DST is currently in effect regionally, otherwise 0. */
+            int dst_local;     /**< 1 if DST is currently in effect in the station's region, otherwise 0. */
+            int dst_scheduled; /**< DST Schedule. 0 means Daylight Savings is not practiced. 1 means U.S./Canada Schedule. 2 means EU Daylight savings schedule. */
+        } local_time;
     };
 };
 /**
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -55,6 +55,8 @@
 #define NRSC5_SAMPLE_RATE_CS16_AM  46511.71875 /**< Sample rate at which nrsc5_pipe_samples_cs16() expects samples (AM only) */
 #define NRSC5_SAMPLE_RATE_AUDIO    44100       /**< Sample rate of outgoing audio */
 
+#define NRSC5_DEVICE_VERSION_LENGTH 4          /**< Length of Core Version & Manufacture Version in SIS Parameter messages. */
+
 #ifdef NRSC5_EXPORTS
 #ifdef __MINGW32__
 #define NRSC5_API __declspec(dllexport)
@@ -188,7 +190,7 @@ enum
     NRSC5_EVENT_AGC,
     NRSC5_EVENT_EXCITER_INFO,
     NRSC5_EVENT_IMPORTER_INFO,
-    NRSC5_EVENT_LEAP_OFFSET,
+    NRSC5_EVENT_LEAP_SECOND_OFFSET,
     NRSC5_EVENT_LOCAL_TIME,
 };
 
@@ -395,9 +397,9 @@ struct nrsc5_event_t
  * - `NRSC5_EVENT_EMERGENCY_ALERT` : emergency alert, see `emergency_alert` member
  * - `NRSC5_EVENT_HERE_IMAGE` : HERE Images traffic/weather map, see `here_image` member
  * - `NRSC5_EVENT_AGC` : automatic gain control status, see `agc` member
- * - `NRSC5_EVENT_EXCITER_INFO` : exciter data, see `device_info` member
- * - `NRSC5_EVENT_IMPORTER_INFO` : importer data, see `device_info` member
- * - `NRSC5_EVENT_LEAP_OFFSET` : leap offset, see `leap_offset` member
+ * - `NRSC5_EVENT_EXCITER_INFO` : exciter data, see `exciter_info` member
+ * - `NRSC5_EVENT_IMPORTER_INFO` : importer data, see `importer_info` member
+ * - `NRSC5_EVENT_LEAP_SECOND_OFFSET` : leap second offset, see `leap_second_offset` member
  * - `NRSC5_EVENT_LOCAL_TIME` : local time data, see `local_time` member
  */
     unsigned int event;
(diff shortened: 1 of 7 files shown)


support/cli.py
```diff
@@ -372,6 +372,28 @@ def callback(self, evt_type, evt):
...
+                          evt.manufacturer_version[3], evt.manufacturer_status)
+        elif evt_type == nrsc5.EventType.LEAP_OFFSET:
+            logging.debug("Leap second offset: pending=%d, current=%d, ALFN adjustment=%d",
+                          evt.pending_leap_offset, evt.current_leap_offset,
+                          evt.alfn_pending_leap_adjustment)
+        elif evt_type == nrsc5.EventType.LOCAL_TIME:
+            logging.debug("Local time: UTC offset=%d minutes, DST scheduled=%d, DST regional? %s, DST local? %s",
+                          evt.utc_offset,
+                          evt.dst_scheduled,
+                          evt.dst_regional if 'yes' else "no", evt.dst_local if 'yes' else 'no')
```
argilo:
```suggestion
                          "yes" if evt.dst_regional else "no", "yes" if evt.dst_local else "no")
```

TheDaChicken: Ah! Thanks for catching that.

include/nrsc5.h
```diff
@@ -185,7 +185,11 @@ enum
     NRSC5_EVENT_HERE_IMAGE,
     NRSC5_EVENT_LOT_HEADER,
     NRSC5_EVENT_LOT_FRAGMENT,
-    NRSC5_EVENT_AGC
+    NRSC5_EVENT_AGC,
+    NRSC5_EVENT_EXCITER_INFO,
+    NRSC5_EVENT_IMPORTER_INFO,
+    NRSC5_EVENT_LEAP_OFFSET,
```
argilo: Maybe it's worth going with `LEAP_SECOND_OFFSET` so that it's clearer what its purpose is without having to dig deeper.

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
             float peak_dbfs;     /**< peak signal amplitude in dB, relative to full scale */
             int is_final;        /**< 1 if this is the final (best) gain value, otherwise 0 */
         } agc;
+        struct {
+            char id[2];                  /**< 2-character identifier of the device, e.g. "GG" or "L7" */
```
argilo: Maybe better to use the name from the spec? (Same below for importer.)
```suggestion
            char id[2];                  /**< Manufacturer ID, e.g. "GG" or "L7" */
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+            int core_version[4];         /**< Core Version number. */
+            int manufacturer_version[4]; /**< Manufacturer-assigned Version number. */
+            int core_status;             /**< Core status. Values such like 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
+            int manufacturer_status;     /**< Manufacturer status. Values such like 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
```
argilo: How about this? (Same below for importer.)
```suggestion
            int core_status;             /**< Core Version status. 0 (Commercial Release), 1 (Engineering Release), or 2 (Patch). */
            int manufacturer_status;     /**< Manufacturer Version status. 0 (Commercial Release), 1 (Engineering Release), or 2 (Patch). */
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+        } importer_info;
+        struct {
+            int pending_leap_offset;                   /**< Future GPS-UTC offset in seconds. Meant to be broadcasted months before the leap seconds and a few hours afterward. */
+            int current_leap_offset;                   /**< Current GPS-UTC offset in seconds. */
+            unsigned int alfn_pending_leap_adjustment; /**< Represents GPS time of a pending leap second adjustment in ALFN.*/
```
```suggestion
            unsigned int alfn_pending_leap_adjustment; /**< ALFN representing the GPS time of a pending leap second adjustment, or 0 if a leap second is not pending */
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+        } leap_offset;
+        struct {
+            int utc_offset;    /**< Local Time Zone UTC Offset in MINUTES. */
```
```suggestion
            int utc_offset;    /**< Local Time Zone UTC Offset in minutes. */
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+            int dst_regional;  /**< 1 if DST is currently in effect regionally, otherwise 0. */
+            int dst_local;     /**< 1 if DST is currently in effect in the station's region, otherwise 0. */
+            int dst_scheduled; /**< DST Schedule. 0 means Daylight Savings is not practiced. 1 means U.S./Canada Schedule. 2 means EU Daylight savings schedule. */
```
```suggestion
            int dst_schedule; /**< DST Schedule. 0 means Daylight Saving Time is not practiced. 1 means U.S./Canada schedule. 2 means EU schedule. */
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+        } importer_info;
+        struct {
+            int pending_leap_offset;                   /**< Future GPS-UTC offset in seconds. Meant to be broadcasted months before the leap seconds and a few hours afterward. */
+            int current_leap_offset;                   /**< Current GPS-UTC offset in seconds. */
+            unsigned int alfn_pending_leap_adjustment; /**< Represents GPS time of a pending leap second adjustment in ALFN.*/
+        } leap_offset;
+        struct {
+            int utc_offset;    /**< Local Time Zone UTC Offset in MINUTES. */
+            int dst_regional;  /**< 1 if DST is currently in effect regionally, otherwise 0. */
+            int dst_local;     /**< 1 if DST is currently in effect in the station's region, otherwise 0. */
```
argilo: The spec says this is a static value that indicates whether DST is practiced locally, not whether it is currently in effect.
```suggestion
            int dst_local;     /**< 1 if DST is practiced locally, otherwise 0. */
```

src/main.c
```diff
@@ -586,6 +586,37 @@ static void callback(const nrsc5_event_t *evt, void *opaque)
...
+    case NRSC5_EVENT_IMPORTER_INFO:
+        log_debug("Importer manuf. \"%c%c\", core version %d.%d.%d.%d, core status %d, manuf. version %d.%d.%d.%d, manuf. status %d",
+                  evt->importer_info.id[0], evt->importer_info.id[1],
+                  evt->importer_info.core_version[0], evt->importer_info.core_version[1], evt->importer_info.core_version[2],
+                  evt->importer_info.core_version[3], evt->importer_info.core_status,
+                  evt->importer_info.manufacturer_version[0], evt->importer_info.manufacturer_version[1], evt->importer_info.manufacturer_version[2],
+                  evt->importer_info.manufacturer_version[3], evt->importer_info.manufacturer_status);
+        break;
+    case NRSC5_EVENT_LEAP_OFFSET:
+        log_debug("Leap second offset: pending=%d, current=%d, ALFN adjustment=%d",
```
argilo:
```suggestion
        log_debug("Leap second offset: pending=%d, current=%d, ALFN of pending adjustment=%d",
```

src/main.c
```diff
@@ -586,6 +586,37 @@ static void callback(const nrsc5_event_t *evt, void *opaque)
...
+                 evt->leap_offset.pending_leap_offset,
+                 evt->leap_offset.current_leap_offset,
+                 evt->leap_offset.alfn_pending_leap_adjustment);
+        break;
+    case NRSC5_EVENT_LOCAL_TIME:
+        log_debug("Local time: UTC offset=%d minutes, DST scheduled=%d, DST regional? %s, DST local? %s",
```
argilo:
```suggestion
        log_debug("Local time: UTC offset=%d minutes, DST schedule=%d, DST in effect regionally? %s, DST practiced locally? %s",
```

include/nrsc5.h
argilo: Also, do you suppose it would be more convenient for API clients (and perhaps more future-proof) if we provided this as a null-terminated string rather than an array?

src/pids.c
```diff
@@ -664,33 +664,52 @@ static void decode_sis(pids_t *st, uint8_t *bits)
...
+                    }
                     break;
                 case 3:
-                    tzo = (parameter >> 5) & 0x7ff;
-                    if (tzo >= 1024) tzo -= 2048;
-                    log_debug("Local time zone offset: %d minutes, DST sched. %d, local DST? %s, regional DST? %s",
-                        tzo, (parameter >> 2) & 0x7, parameter & 0x2 ? "yes" : "no", parameter & 0x1 ? "yes" : "no");
+                    dst_regional = st->parameters[3] & 0x1;
+                    dst_local = (st->parameters[3] >> 1) & 0x1;
+                    dst_scheduled = (st->parameters[3] >> 2) & 0x7;
```
argilo:
```suggestion
                    dst_schedule = (st->parameters[3] >> 2) & 0x7;
```

src/private.h
```diff
@@ -87,3 +87,11 @@ void nrsc5_report_emergency_alert(nrsc5_t *st, const char *message, const uint8_
...
                              float latitude1, float longitude1, float latitude2, float longitude2,
                              const char *name, unsigned int size, const uint8_t *data);
+void nrsc5_report_exciter_info(nrsc5_t *st, const char id[2], const int core_version[4], const int manufacturer_version[4],
+                               int core_status, int manufacturer_status, int importer_connected);
+void nrsc5_report_importer_info(nrsc5_t *st, const char id[2], const int core_version[4], const int manufacturer_version[4],
+                                int core_status, int manufacturer_status);
+void nrsc5_report_leap(nrsc5_t *st, int pending_leap_offset, int current_leap_offset,
+                       unsigned int alfn_leap_adjustment);
+void nrsc5_report_local_time(nrsc5_t *st, int utc_offset, int dst_regional, int dst_local,
+                             int dst_schedule);
```
argilo: Nitpick: missing carriage return at the end of the file

include/nrsc5.h
argilo: Maybe `manufacturer_id` would be clearer as a field name as well.

support/cli.py
```diff
@@ -372,6 +372,28 @@ def callback(self, evt_type, evt):
             logging.info("HERE Image: type=%s, seq=%d, n1=%d, n2=%d, time=%s, lat1=%.5f, lon1=%.5f, lat2=%.5f, lon2=%.5f, name=%s, size=%d",
                          evt.image_type.name, evt.seq, evt.n1, evt.n2, time_str, evt.latitude1, evt.longitude1,
                          evt.latitude2, evt.longitude2, evt.name, len(evt.data))
+        elif evt_type == nrsc5.EventType.EXCITER_INFO:
+            logging.debug("Exciter manuf. \"%c%c\", core version %d.%d.%d.%d, core status %d, manuf. version %d.%d.%d.%d, manuf. status %d, importer connected? %s",
+                          evt.id[0], evt.id[1], evt.core_version[0], evt.core_version[1], evt.core_version[2],
```
argilo: An array of characters gets even weirder in Python, so I would think a string is the way to go.

include/nrsc5.h
TheDaChicken: That's a good question. I do think `manufacturer_id` might be better especially fits with the comment "clearer what its purpose is without having to dig deeper"

TheDaChicken: I think I can see that it make senses to use a null-terminated string especially since its what other structs for the whole API. I will change it to that.

support/cli.py
TheDaChicken: Less pythonic, yeah

include/nrsc5.h
TheDaChicken: That seems kind of long. But I understand. That would mean that there won't be any fields other than seconds right?

argilo: Right, messages 0-2 all pertain to leap seconds and nothing else.

argilo: Basically my thinking was that "leap" on its own doesn't convey what it's about, but "leap second" should be unambiguous.

TheDaChicken: That's true I think. You hear often "leap seconds" not "leap." This may not imply completely that all the fields must be seconds.

include/nrsc5.h
```diff
@@ -185,7 +185,11 @@ enum
     NRSC5_EVENT_HERE_IMAGE,
     NRSC5_EVENT_LOT_HEADER,
     NRSC5_EVENT_LOT_FRAGMENT,
-    NRSC5_EVENT_AGC
+    NRSC5_EVENT_AGC,
+    NRSC5_EVENT_EXCITER_INFO,
+    NRSC5_EVENT_IMPORTER_INFO,
+    NRSC5_EVENT_LEAP_SECONDS_OFFSET,
```
```suggestion
    NRSC5_EVENT_LEAP_SECOND_OFFSET,
```

include/nrsc5.h
```diff
@@ -391,6 +395,10 @@ struct nrsc5_event_t
  * - `NRSC5_EVENT_EMERGENCY_ALERT` : emergency alert, see `emergency_alert` member
  * - `NRSC5_EVENT_HERE_IMAGE` : HERE Images traffic/weather map, see `here_image` member
  * - `NRSC5_EVENT_AGC` : automatic gain control status, see `agc` member
+ * - `NRSC5_EVENT_EXCITER_INFO` : exciter data, see `exciter_info` member
+ * - `NRSC5_EVENT_IMPORTER_INFO` : importer data, see `importer_info` member
+ * - `NRSC5_EVENT_LEAP_SECOND_OFFSET` : leap offset, see `leap_offset` member
```
```suggestion
 * - `NRSC5_EVENT_LEAP_SECOND_OFFSET` : leap second offset, see `leap_offset` member
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
             float peak_dbfs;     /**< peak signal amplitude in dB, relative to full scale */
             int is_final;        /**< 1 if this is the final (best) gain value, otherwise 0 */
         } agc;
+        struct {
+            const char* manufacturer_id; /**< Manufacturer ID, e.g. "GG" or "L7" */
+            int core_version[4];         /**< Core Version number. */
+            int manufacturer_version[4]; /**< Manufacturer-assigned Version number. */
+            int core_status;             /**< Core Version status. 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
+            int manufacturer_status;     /**< Manufacturer Version status. 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
```
argilo: Maybe it would look a bit nicer to group the core version & manufacturer version fields. (Same below for importer.)
```suggestion
            int core_version[4];         /**< Core Version number. */
            int core_status;             /**< Core Version status. 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
            int manufacturer_version[4]; /**< Manufacturer-assigned Version number. */
            int manufacturer_status;     /**< Manufacturer Version status. 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+        } importer_info;
+        struct {
+            int pending_leap_offset;                   /**< Future GPS-UTC offset in seconds. Meant to be broadcasted months before the leap seconds and a few hours afterward. */
+            int current_leap_offset;                   /**< Current GPS-UTC offset in seconds. */
```
argilo: We could perhaps shorten these names. The standard calls them "Pending Offset" and "Current Offset", and "leap" is already in the struct name.
```suggestion
            int pending_offset;                   /**< Future GPS-UTC offset in seconds. Meant to be broadcasted months before the leap seconds and a few hours afterward. */
            int current_offset;                   /**< Current GPS-UTC offset in seconds. */
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+            unsigned int alfn_pending_leap_adjustment; /**< ALFN representing the GPS time of a pending leap second adjustment, or 0 if a leap second is not pending.*/
```
argilo: `pending_alfn` might pair well with `pending_offset`.
```suggestion
            unsigned int pending_alfn; /**< ALFN representing the GPS time of a pending leap second adjustment, or 0 if a leap second is not pending.*/
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+        } leap_offset;
+        struct {
+            int utc_offset;    /**< Local Time Zone UTC Offset in minutes. */
+            int dst_regional;  /**< 1 if DST is currently in effect regionally, otherwise 0. */
+            int dst_local;     /**< 1 if DST is practiced locally, otherwise 0. */
+            int dst_schedule;  /**< DST Schedule. 0 means Daylight Saving Time is not practiced. 1 means U.S./Canada schedule. 2 means EU Daylight savings schedule. */
```
```suggestion
            int dst_schedule;  /**< DST Schedule. 0 means Daylight Saving Time is not practiced. 1 means U.S./Canada schedule. 2 means EU schedule. */
```

src/pids.c
```diff
@@ -664,33 +664,53 @@ static void decode_sis(pids_t *st, uint8_t *bits)
...
                 case 2:
-                    if (st->parameters[1] >= 0 && st->parameters[2] >= 0)
-                        log_debug("ALFN of pending leap second adjustment: %d", (unsigned int)st->parameters[2] << 16 | st->parameters[1]);
+                    if (st->parameters[0] >= 0 && st->parameters[1] >= 0 && st->parameters[2] >= 0)
+                    {
+                        const int pending_leap_offset = st->parameters[0] >> 8;
+                        const int current_leap_offset = st->parameters[0] & 0xff;
+                        const unsigned int alfn_pending_leap_adjustment = (unsigned int)st->parameters[2] << 16 | st->parameters[1];
+
+                        nrsc5_report_leap(st->input->radio, pending_leap_offset, current_leap_offset, alfn_pending_leap_adjustment);
```
argilo:
```suggestion
                        const int pending_offset = st->parameters[0] >> 8;
                        const int current_offset = st->parameters[0] & 0xff;
                        const unsigned int pending_alfn = (unsigned int)st->parameters[2] << 16 | st->parameters[1];

                        nrsc5_report_leap(st->input->radio, pending_offset, current_offset, pending_alfn);
```

include/nrsc5.h
```diff
@@ -573,6 +581,32 @@ struct nrsc5_event_t
...
+            int core_version[4];         /**< Core Version number. */
+            int manufacturer_version[4]; /**< Manufacturer-assigned Version number. */
+            int core_status;             /**< Core Version status. 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
+            int manufacturer_status;     /**< Manufacturer Version status. 0 (Commercial Release), 1 (Engineering Release), 2 (Patch). */
+        } importer_info;
+        struct {
+            int pending_leap_offset;                   /**< Future GPS-UTC offset in seconds. Meant to be broadcasted months before the leap seconds and a few hours afterward. */
+            int current_leap_offset;                   /**< Current GPS-UTC offset in seconds. */
+            unsigned int alfn_pending_leap_adjustment; /**< ALFN representing the GPS time of a pending leap second adjustment, or 0 if a leap second is not pending.*/
+        } leap_offset;
```
argilo: Maybe for consistency with the event name, we could use `leap_second_offset` for the struct name. It doesn't seem too unwieldy, since it's only used in a handful of places.
```suggestion
        } leap_second_offset;
```

include/nrsc5.h
TheDaChicken: It does seem to look nicer.

include/nrsc5.h
TheDaChicken: This also shortens things a lot on the Python side too which is nice.

include/nrsc5.h
TheDaChicken: Yeah, that works. It is missing the adjustment part but it's shorten & 'adjustment' part is documented.

support/nrsc5.py
```diff
@@ -54,6 +54,7 @@ class EventType(enum.Enum):
 SAMPLE_RATE_CS16_AM = 46511.71875
 SAMPLE_RATE_AUDIO = 44100
 
+NRSC5_DEVICE_VERSION_LENGTH = 4
```
argilo: No need to prefix `NRSC5_` here, since Python modules are namespaced.

TheDaChicken: Ah! Yep.

support/nrsc5.py
```diff
@@ -224,7 +229,10 @@ class PacketFlags(enum.IntFlag):
 HEREImage = collections.namedtuple("HEREImage", ["image_type", "seq", "n1", "n2", "time_utc", "latitude1", "longitude1",
                                                  "latitude2", "longitude2", "name", "data"])
 AGC = collections.namedtuple("AGC", ["gain_db", "peak_dbfs", "is_final"])
-
+ExciterInfo = collections.namedtuple("ExciterInfo", ["manufacturer_id", "core_version", "manufacturer_version", "core_status", "manufacturer_status", "importer_connected"])
+ImporterInfo = collections.namedtuple("ImporterInfo", ["manufacturer_id", "core_version", "manufacturer_version", "core_status", "manufacturer_status"])
```
argilo: These tuples should probably be reordered to match the ordering of the C structs.

support/nrsc5.py
```diff
@@ -584,6 +592,39 @@ class _AGC(ctypes.Structure):
         ("is_final", ctypes.c_int),
     ]
 
+class _ExciterInfo(ctypes.Structure):
+    _fields_ = [
+        ("manufacturer_id", ctypes.c_char_p),
+        ("core_version", ctypes.c_int * DEVICE_VERSION_LENGTH),
+        ("manufacturer_version", ctypes.c_int * DEVICE_VERSION_LENGTH),
+        ("core_status", ctypes.c_int),
+        ("manufacturer_status", ctypes.c_int),
```
argilo, line 601: Field order looks incorrect here. Same below for importer.

TheDaChicken: Ah! Thanks for catching that!

support/nrsc5.py
TheDaChicken: Yep

PR #492, opened 2026-04-15
Refactor SIS

TheDaChicken:
The `decode_sis` function has gotten really big. This function may get bigger as time goes on. This is to split up sis into a function for each sis message that independently does its own work.

Plus, NRSC5-E defines a new PIDS type called LLDS. This PR also adds a library debug message for LLDS to identify if a station is broadcasting this type. 

I have done my own testing to confirm that there is no segment faults on my recordings.

### DIFF
--- a/src/pids.c
+++ b/src/pids.c
@@ -35,6 +35,14 @@
 #define SIS_MSG_ID_EMERGENCY_ALERTS_MESSAGE 9
 #define SIS_MSG_ID_ADV_SERVICE_INFORMATION 10
 
+#define SIS_SERVICE_CATEGORY_AUDIO 0
+#define SIS_SERVICE_CATEGORY_DATA 1
+
+#define SIS_EMERGENCY_ALERTS_SAME 0
+#define SIS_EMERGENCY_ALERTS_FIPS 1
+#define SIS_EMERGENCY_ALERTS_ZIP  2
+
+
 static char *chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ ?-*$ ";
 static int payload_sizes[] = {
     32, 22, 58, 32, 27, 58, 27, 22,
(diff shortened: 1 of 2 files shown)


src/pids.c
```diff
@@ -382,537 +394,628 @@ static void reset_alert(pids_t *st)
...
+    int updated = 0, off = 0, j;
+    int prog_num;
+    asd_t audio_service;
+    dsd_t data_service;
+
+    const unsigned int category = decode_int(bits, &off, 2);
+
+    switch (category)
+    {
+    case 0:
```
argilo: Perhaps we could use a macro for these numbers, like `SIS_SERVICE_CATEGORY_AUDIO` and `SIS_SERVICE_CATEGORY_DATA`.

src/pids.c
```diff
@@ -382,537 +394,628 @@ static void reset_alert(pids_t *st)
...
+            if (sis_decode_universal_short_station_name(st, bits + off))
+                updated = 1;
+            off += 58;
+            break;
+        case SIS_MSG_ID_EMERGENCY_ALERTS_MESSAGE:
+            if (sis_decode_emergency_alerts(st, bits + off))
+                updated = 1;
+            off += 58;
+            break;
+        default:
```
argilo, line 1015: Perhaps we should log something in the default case, like `log_warn("Unknown SIS MSG ID: %d", msg_id);`.

I tested this, and didn't get any hits with my suite of recordings.

src/pids.c
```diff
@@ -924,21 +1027,29 @@ static void decode_sis(pids_t *st, uint8_t *bits)
...
+    }
+
+    if (check_crc12(pids))
+    {
+        const int type = pids[0];
+
+        if (type == PIDS_TYPE_SIS)
+            sis_decode(st, pids + 1);
+        else if (type == PIDS_TYPE_LLDS)
+            log_debug("Ignoring LLDS frame.");
```
argilo:
```suggestion
            log_debug("Ignoring LLDS frame");
```

src/pids.c
```diff
@@ -21,6 +21,27 @@
...
+#define SIS_MSG_ID_UNIVERSAL_SHORT_STATION_NAME 8
+#define SIS_MSG_ID_EMERGENCY_ALERTS_MESSAGE 9
+#define SIS_MSG_ID_ADV_SERVICE_INFORMATION 10
+
+#define SIS_SERVICE_CATEGORY_AUDIO 0
+#define SIS_SERVICE_CATEGORY_DATA 1
+
+#define SIS_EMERGENCY_ALERTS_SAME 0
+#define SIS_EMERGENCY_ALERTS_FIPS 1
+#define SIS_EMERGENCY_ALERTS_ZIP  2
```
argilo: We could maybe be more specific with these names.
```suggestion
#define SIS_EA_LOCATION_FORMAT_SAME 0
#define SIS_EA_LOCATION_FORMAT_FIPS 1
#define SIS_EA_LOCATION_FORMAT_ZIP  2
```

PR #500, opened 2026-04-19
Refactor audio output queue

### LINKED ISSUE #495: Dropped audio samples when playing back recording in C CLI
Sometimes when playing back an I/Q recording in the C CLI, I see a warning about dropped audio samples during startup:
```
$ src/nrsc5 --am --iq-input-format cs16 -r ~/Documents/nrsc5-am/chlo.cs16 0
00:12:11 Synchronized
00:12:11 Frequency offset: 0 Hz
00:12:11 Primary service mode: 1
00:12:11 Digital bandwidth: reduced
00:12:11 Station name: CHLO
00:12:11 Country: CA, FCC facility ID: 106051
00:12:11 Audio program 0: public, type: Talk, sound experience 0
00:12:11 Station location: 43.5880, -79.8838, 0m
00:12:11 Slogan: Multicultural Radio
00:12:11 Audio service 0: public, type: Talk, codec: 2, blend: 2, gain: 3 dB, delay: 144, latency: 8
00:12:11 BER: 0.012153, avg: 0.012153, min: 0.012153, max: 0.012153
00:12:11 Audio bit rate: 18.7 kbps
00:12:11 BER: 0.012736, avg: 0.012444, min: 0.012153, max: 0.012736
00:12:11 Audio bit rate: 18.5 kbps
00:12:11 BER: 0.011833, avg: 0.012241, min: 0.011833, max: 0.012736
00:12:11 Audio bit rate: 18.8 kbps
00:12:11 BER: 0.011778, avg: 0.012125, min: 0.011778, max: 0.012736
00:12:11 Audio output timed out, dropping samples
00:12:11 Audio bit rate: 18.5 kbps
00:12:11 BER: 0.012194, avg: 0.012139, min: 0.011778, max: 0.012736
00:12:11 Audio bit rate: 18.7 kbps
00:12:11 BER: 0.012375, avg: 0.012178, min: 0.011778, max: 0.012736
00:12:11 Audio bit rate: 18.3 kbps
```
I don't yet know why this occurs, but there's no good reason for audio samples to be dropped while playing back an I/Q file, since there's no pressure to keep up with a live incoming sample stream (except perhaps when reading from stdin or a fifo).

### LINKED ISSUE #497: Excessively large audio buffer in C CLI
nrsc5's C CLI uses a 128-packet (= ~6 second) audio buffer, which seems excessive.

In #399 the size of the audio buffer in the Python CLI was reduced from 64 to 16. It would probably be sensible to reduce the C CLI's audio buffer to the same size.

argilo:
Fixes #495.
Fixes #497.

Here I have fixed several issues related to audio output in the C and Python CLI applications:

1. In the Python CLI, when receiving from an SDR and playing back audio on a sound card, clock drift in the sound card could eventually lead to the audio output queue becoming full. At this point, the libnrsc5 callback function would be blocked, resulting in dropped I/Q samples and loss of synchronization. The C CLI avoids this problem by dropping audio samples when the queue becomes full, and I've implemented that approach in the Python CLI as well.
2. The C CLI had an excessively large audio output queue (128 packets, or ~6 seconds). I've reduced it to 16 packets, matching the Python CLI.
3. To compensate for the (former) lack of elastic buffering in libnrsc5, the C CLI waited until the audio queue reached 8 packets before beginning audio output. This approach is no longer necessary since elastic buffering was implemented in #399, so I have removed it.
4. In the C CLI, the `push_audio_buffer` function used a 100-ms timeout to avoid dropping audio output when reading from a file. This does not work correctly, as explained in #495. Instead of the timeout, I've changed the code so that it blocks indefinitely while waiting for space in the output queue (if reading from an I/Q file) or aborts immediately and drops the audio packet (if receiving from an SDR).
5. In the C CLI, the `push_audio_buffer` function briefly unlocked its mutex while performing a `memcpy`. After re-locking, it was necessary to re-check the value of `st->program` in case it had changed. Since the `memcpy` should be fast, I've changed the code to leave the mutex locked.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -18,7 +18,6 @@
 #include <math.h>
 #include <nrsc5.h>
 #include <pthread.h>
-#include <sys/time.h>
 #include <unistd.h>
 
 #include <assert.h>
@@ -42,8 +41,7 @@
 #include "bitwriter.h"
 #include "log.h"
 
-#define AUDIO_BUFFERS 128
-#define AUDIO_THRESHOLD 8
+#define AUDIO_BUFFERS 16
 #define AUDIO_DATA_LENGTH 8192
 #define FILE_BUFFER_LENGTH 32768
 #define STDIN_POLL_RATE_MS 100
@@ -80,7 +78,6 @@ typedef struct {
     pthread_cond_t cond;
 
     unsigned int program;
-    unsigned int audio_ready;
     unsigned int audio_packets_valid;
     unsigned int audio_packets;
     unsigned int audio_bytes;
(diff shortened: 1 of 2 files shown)


argilo:
To test the behaviour when the audio queue fills up, I modified the CLI apps' audio output threads to play back a small amount of extra audio samples beyond the ones arriving on the queue.

argilo:
Prior to this change, the C CLI would drop 128 audio packets (6 seconds) if its queue filled up. Now it drops only a single packet.

argilo:
Now that the 8-packet startup delay is gone, switching between audio subchannels using the keyboard is almost instantaneous. I'm liking that!

TheDaChicken:
I found this to hugely improve playing iq recordings and live decoding while doing other things in the background. 

The small pauses due to dropped samples were less annoying than having no audio for 6 seconds.

Issue #457, opened 2025-10-17
Get rid of dynamic memory allocations in libnrsc5?


argilo:
libnrsc5 allocates memory using `malloc` in various places. For instance, the convolutional decoder allocates a relatively large trellis object at the start of each decode, and frees it when finished. Each memory allocation opens up the possibility of an allocation failure, and in many cases libnrsc5 fails to check the return value of `malloc`.

Since most things in the NRSC-5 standard are of fixed or bounded size, I expect that hardware receivers use a static memory layout. It may be worth taking this approach in libnrsc5 as well.

TheDaChicken:
Just double-checking. I currently cannot think of any other better way. 

When using callbacks on the API, I have to copy memory into my own variables since I cannot take ownership of the memory. That is completely OK. This may make more sense if the memory was stack allocated, copying would be the only correct option.

PR #501, opened 2026-04-19
Refactor Decoder

TheDaChicken:
~~Fixes an issue raised in #457~~ (to be moved in a different pr)

While working on the API for the local_time SIS parameter, I wanted to try out decoding the ALFN to experiment with what stations are sending out. That requires the current block count to construct ALFN. 
The current block count isn't directly associated with each PIDS. To make this easier, I decided to refactor the decoder to make the current block count more accessible.

This is the improvements:

- `decode_set_block` is removed since block count is now provided with each call. 
- For any decoding changes based on PSMI can be changed through `decode_set_psmi` now. `decode_set_px1_length` is not needed. 
- Each deinterleaver is now into their own individual functions. This can make deinterleaving a lot easier for advanced service modes. This can also the possibly to deinterleave per block instead of by a complete frame a bit easier. 
- Constants for the convolutional decoder of P, E1, E2 
- ~~The convolutional decoder's path and trellis is now separated. The possibly for stack allocation is easier.~~ (to be moved in a different pr)
- ~~The convolutional decoder's path and trellis is now allocated upfront before decoding and freed after decoding is finished. This does make memory usage (\~18.7 MB) from convolutional decoder noticeable. For example, Task Manager on Windows now shows the memory usage from the convolutional decoder.~~ (to be moved in a different pr)

There is still some code duplication but good enough for what it is at the moment.

I did learn a lot by asking @argilo many questions to make sure this was refactored properly.

### DIFF
--- a/src/decode.c
+++ b/src/decode.c
@@ -525,7 +525,7 @@ void decode_process_p1_p3_am(decode_t *st, const unsigned int bc)
                 if (st->input->sync.psmi != SERVICE_MODE_MA3)
                 {
                     total_frame_length += P3_FRAME_LEN_ENCODED_MA1;
-                    nrsc5_conv_decode_e1(st->viterbi_p3_am, st->scrambler_p3_am, P3_FRAME_LEN_MA1);
+                    nrsc5_conv_decode_e2_e3(st->viterbi_p3_am, st->scrambler_p3_am, P3_FRAME_LEN_MA1);
                     st->am_errors += bit_errors_e2(st->viterbi_p3_am, st->scrambler_p3_am, P3_FRAME_LEN_MA1);
                     descramble(st->scrambler_p3_am, P3_FRAME_LEN_MA1);
                     frame_push(&st->input->frame, st->scrambler_p3_am, P3_FRAME_LEN_MA1, P3_LOGICAL_CHANNEL);
--- a/src/decode.c
+++ b/src/decode.c
@@ -22,21 +22,50 @@
 #include "private.h"
 
 /* 1012s.pdf figure 10-4 */
-static int bl_delay[] = { 2, 1, 5 };
-static int ml_delay[] = { 11, 6, 7 };
-static int bu_delay[] = { 10, 8, 9 };
-static int mu_delay[] = { 4, 3, 0 };
-static int el_delay[] = { 0, 1 };
-static int eu_delay[] = { 2, 3, 5, 4 };
+static const int bl_delay[] = { 2, 1, 5 };
+static const int ml_delay[] = { 11, 6, 7 };
+static const int bu_delay[] = { 10, 8, 9 };
+static const int mu_delay[] = { 4, 3, 0 };
+static const int el_delay[] = { 0, 1 };
+static const int eu_delay[] = { 2, 3, 5, 4 };
+
+static const int8_t V_PM[] = {
+    10, 2, 18, 6, 14, 8, 16, 0, 12, 4,
+    11, 3, 19, 7, 15, 9, 17, 1, 13, 5
+};
+
+static const struct lte_conv_code conv_code_p = {
+    .n = 3,
+    .k = 7,
+    .len = P1_FRAME_LEN_FM,
+    .gen = { 0133, 0171, 0165 },
+    .term = CONV_TERM_TAIL_BITING,
+};
+
+static const struct lte_conv_code conv_code_e1 = {
+    .n = 3,
+    .k = 9,
+    .len = P3_FRAME_LEN_MA3,
+    .gen = { 0561, 0657, 0711 },
+    .term = CONV_TERM_TAIL_BITING,
+};
+
+static const struct lte_conv_code conv_code_e2_e3 = {
+    .n = 3,
+    .k = 9,
+    .len = P3_FRAME_LEN_MA1,
+    .gen = { 0561, 0753, 0711 },
+    .term = CONV_TERM_TAIL_BITING,
+};
 
 /* 1012s.pdf figure 10-5 */
-static int pids_il_delay[] = { 0, 1, 12, 13, 6, 5, 18, 17, 11, 7, 23, 19 };
-static int pids_iu_delay[] = { 2, 4, 14, 16, 3, 8, 15, 20, 9, 10, 21, 22 };
+static const int pids_il_delay[] = { 0, 1, 12, 13, 6, 5, 18, 17, 11, 7, 23, 19 };
+static const int pids_iu_delay[] = { 2, 4, 14, 16, 3, 8, 15, 20, 9, 10, 21, 22 };
 
-static int bit_map(unsigned char matrix[PARTITION_WIDTH_AM * BLKSZ * 8], int b, int k, int p)
+static int bit_map(const unsigned char matrix[PARTITION_WIDTH_AM * BLKSZ * 8], const int b, const int k, const int p)
 {
-    int col = (9*k) % 25;
-    int row = (11*col + 16*(k/25) + 11*(k/50)) % 32;
+    const int col = (9*k) % 25;
+    const int row = (11*col + 16*(k/25) + 11*(k/50)) % 32;
     return (matrix[PARTITION_WIDTH_AM * (b*BLKSZ + row) + col] >> p) & 1;
 }
 
@@ -200,9 +229,9 @@ static void interleaver_ma1(decode_t *st)
 }
 
 // calculate number of bit errors by re-encoding and comparing to the input
-static int bit_errors(int8_t *coded, uint8_t *decoded, unsigned int k, unsigned int frame_len,
-                      unsigned int g1, unsigned int g2, unsigned int g3,
-                      uint8_t *puncture, int puncture_len)
+static int bit_errors(int8_t *coded, uint8_t *decoded, const unsigned int k, unsigned int frame_len,
+                      const unsigned int gens[3],
+                      const uint8_t *puncture, const int puncture_len)
 {
     uint16_t r = 0;
     unsigned int i, j, errors = 0;
@@ -216,39 +245,33 @@ static int bit_errors(int8_t *coded, uint8_t *decoded, unsigned int k, unsigned 
         // shift in new bit
         r = (r >> 1) | (decoded[i] << (k-1));
 
-        if (puncture[j % puncture_len] && ((coded[j] > 0) != __builtin_parity(r & g1)))
+        if (puncture[j % puncture_len] && ((coded[j] > 0) != __builtin_parity(r & gens[0])))
             errors++;
-        if (puncture[(j+1) % puncture_len] && ((coded[j+1] > 0) != __builtin_parity(r & g2)))
+        if (puncture[(j+1) % puncture_len] && ((coded[j+1] > 0) != __builtin_parity(r & gens[1])))
             errors++;
-        if (puncture[(j+2) % puncture_len] && ((coded[j+2] > 0) != __builtin_parity(r & g3)))
+        if (puncture[(j+2) % puncture_len] && ((coded[j+2] > 0) != __builtin_parity(r & gens[2])))
             errors++;
     }
 
     return errors;
 }
 
-static int bit_errors_p1_fm(int8_t *coded, uint8_t *decoded)
+static int bit_errors_p_fm(int8_t *coded, uint8_t *decoded, const int len)
 {
-    uint8_t puncture[] = {1, 1, 1, 1, 1, 0};
-    return bit_errors(coded, decoded, 7, P1_FRAME_LEN_FM, 0133, 0171, 0165, puncture, 6);
+    const uint8_t puncture[] = {1, 1, 1, 1, 1, 0};
+    return bit_errors(coded, decoded, conv_code_p.k, len, conv_code_p.gen, puncture, 6);
 }
 
-static int bit_errors_p1_am(int8_t *coded, uint8_t *decoded)
+static int bit_errors_e1(int8_t *coded, uint8_t *decoded, const int len)
 {
-    uint8_t puncture[] = {1, 0, 1, 1, 0, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1};
-    return bit_errors(coded, decoded, 9, P1_FRAME_LEN_AM, 0561, 0657, 0711, puncture, 15);
+    const uint8_t puncture[] = {1, 0, 1, 1, 0, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1};
+    return bit_errors(coded, decoded, conv_code_e1.k, len, conv_code_e1.gen, puncture, 15);
 }
 
-static int bit_errors_p3_ma1(int8_t *coded, uint8_t *decoded)
+static int bit_errors_e2(int8_t *coded, uint8_t *decoded, const int len)
 {
-    uint8_t puncture[] = {1, 0, 1, 1, 0, 0};
-    return bit_errors(coded, decoded, 9, P3_FRAME_LEN_MA1, 0561, 0753, 0711, puncture, 6);
-}
-
-static int bit_errors_p3_ma3(int8_t *coded, uint8_t *decoded)
-{
-    uint8_t puncture[] = {1, 0, 1, 1, 0, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1};
-    return bit_errors(coded, decoded, 9, P3_FRAME_LEN_MA3, 0561, 0657, 0711, puncture, 15);
+    const uint8_t puncture[] = {1, 0, 1, 1, 0, 0};
+    return bit_errors(coded, decoded, conv_code_e2_e3.k, len, conv_code_e2_e3.gen, puncture, 6);
 }
 
 static void descramble(uint8_t *buf, unsigned int length)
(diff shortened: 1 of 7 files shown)


src/defines.h
```diff
@@ -70,6 +71,10 @@
 #define ELASTIC_BUFFER_LEN 64
 // number of subcarriers per AM partition
 #define PARTITION_WIDTH_AM 25
+// number of subcarriers per FM partition
+#define PARTITION_WIDTH_FM 19
+// PCM block size
+#define PCM_BLOCK_SIZE (2 * 20 * (PARTITION_WIDTH_FM - 1) * BLKSZ)
```
argilo: The `PARTITION_DATA_CARRIERS` macro (currently in sync.c) might be more appropriate than `PARTITION_WIDTH_FM - 1`.

argilo: What does `PCM` mean?

TheDaChicken: I didn't know a good name for the length of data that gets pushed into `decode_push_pm`. Is there a better name?

argilo: `PM_BLOCK_SIZE` then? I just didn't understand where the C came from.

TheDaChicken: OH! Yes, that's what I meant. Should I use that?

argilo: The 20 could perhaps be replaced with a macro like `PRIMARY_MAIN_PARTITIONS`.

argilo: Actually I see sync.c already has `#define PM_PARTITIONS 10` so you could use that and multiply by 2 for the number of sidebands.

src/decode.h
```diff
@@ -24,13 +24,10 @@ typedef struct
     int8_t buffer_pm[720 * BLKSZ * 16];
```
argilo: Replace `720 * BLKSZ` with the new `PM_BLOCK_SIZE` macro?

src/sync.c
```diff
@@ -506,97 +504,110 @@ void sync_process_fm(sync_t *st)
...
+        const float mer_lb = 2.0f * BLKSZ * (float)(partitions_per_band * PARTITION_DATA_CARRIERS) / error_lb;
+        const float mer_ub = 2.0f * BLKSZ * (float)(partitions_per_band * PARTITION_DATA_CARRIERS) / error_ub;
+        const float mult_lb = fmaxf(fminf(mer_lb * 10, 127), 1);
+        const float mult_ub = fmaxf(fminf(mer_ub * 10, 127), 1);
 
-        decode_set_block(&st->input->decode, st->bc);
+        int8_t buffer_pm[PCM_BLOCK_SIZE];
+        int8_t buffer_px1[P3_FRAME_LEN_MP3_MP11];
+        int8_t buffer_px2[P3_FRAME_LEN_MP3_MP11];
+        int out_px1 = 0, out_px2 = 0, out_pm = 0;
```
argilo: Nitpick: May as well keep these in a consistent order.
```suggestion
        int out_pm = 0, out_px1 = 0, out_px2 = 0;
```

src/decode.c
```diff
@@ -15,28 +15,59 @@
...
 #include <string.h>
 
-#include "conv.h"
 #include "decode.h"
 #include "input.h"
 #include "pids.h"
 #include "private.h"
 
+#define CONV_K_AM 9
+#define CONV_K_FM 7
```
argilo: This isn't really an FM vs AM thing. The new FM advanced service modes use constraint length 9.

Maybe we don't need macros for the K values.

src/decode.c
```diff
@@ -200,9 +231,9 @@ static void interleaver_ma1(decode_t *st)
 }
 
 // calculate number of bit errors by re-encoding and comparing to the input
-static int bit_errors(int8_t *coded, uint8_t *decoded, unsigned int k, unsigned int frame_len,
-                      unsigned int g1, unsigned int g2, unsigned int g3,
-                      uint8_t *puncture, int puncture_len)
+static int bit_errors(int8_t *coded, uint8_t *decoded, const unsigned int k, unsigned int frame_len,
+                      const unsigned int gens[0],
```
argilo: This array should have length 3, shouldn't it?

TheDaChicken: I put a zero???? How did the compiler allow this??

src/decode.c
```diff
@@ -15,28 +15,59 @@
...
+
+static const struct lte_conv_code conv_code_e1 = {
+    .n = 3,
+    .k = CONV_K_AM,
+    .len = P3_FRAME_LEN_MA3,
+    .gen = { 0561, 0657, 0711 },
+    .term = CONV_TERM_TAIL_BITING,
+};
+
+static const struct lte_conv_code conv_code_e2 = {
```
argilo: I guess `e2` is now being used for the E3 convolutional decoder as well, since they differ only in puncturing? If so, it might make sense to change the names of all the `_e2` things to `_e2_e3` (except `bit_errors_e2` which uses the E2-specific puncturing).

src/decode.c
argilo: Well, whatever you put it decays to a simple pointer. The number is just for the programmer's convenience.

TheDaChicken: That's silly.

src/defines.h
```diff
@@ -70,6 +71,14 @@
 #define ELASTIC_BUFFER_LEN 64
 // number of subcarriers per AM partition
 #define PARTITION_WIDTH_AM 25
+// number of subcarriers per FM partition
+#define PARTITION_WIDTH_FM 19
+// number of data carries per FM partition
```
argilo, line 76:
```suggestion
// number of data subcarriers per FM partition
```

src/defines.h
```diff
@@ -70,6 +71,14 @@
...
+#define PARTITION_DATA_CARRIERS 18
+// number of partitions on PM
```
argilo:
```suggestion
// number of partitions in each Primary Main (PM) sideband
```

src/defines.h
```diff
@@ -70,6 +71,14 @@
...
+#define PM_PARTITIONS 10
+// PM block size
```
argilo:
```suggestion
// size of one block in the PM interleaver matrix
```

src/defines.h
```diff
@@ -49,7 +49,8 @@
 #define PIDS_FRAME_LEN_ENCODED_FM (PIDS_FRAME_LEN * 5 / 2)
 #define PIDS_FRAME_LEN_ENCODED_AM (PIDS_FRAME_LEN * 3)
 // bits per P3 frame
-#define P3_FRAME_LEN_FM 4608
+#define P3_FRAME_LEN_MP3_MP11 4608
+#define P3_FRAME_LEN_MP2 (P3_FRAME_LEN_MP3_MP11 / 2)
```
argilo: It might make sense to define MP2's P3 size directly, since the number is listed separately in the spec and not linked to MP3.
```suggestion
#define P3_FRAME_LEN_MP2 2304
#define P3_FRAME_LEN_MP3_MP11 4608
```

src/decode.c
```diff
@@ -22,21 +22,50 @@
...
+static const int mu_delay[] = { 4, 3, 0 };
+static const int el_delay[] = { 0, 1 };
+static const int eu_delay[] = { 2, 3, 5, 4 };
+
+static const int8_t V_PM[] = {
+    10, 2, 18, 6, 14, 8, 16, 0, 12, 4,
+    11, 3, 19, 7, 15, 9, 17, 1, 13, 5
+};
+
+static const struct lte_conv_code conv_code_p = {
```
argilo: I'm not sure what `p` means here. The two mother codes used in FM differ in their K values, so maybe `k7` for the K=7 mother code would make sense.
```suggestion
static const struct lte_conv_code conv_code_k7 = {
```

src/decode.c
```diff
@@ -216,39 +245,33 @@ static int bit_errors(int8_t *coded, uint8_t *decoded, unsigned int k, unsigned
...
-        if (puncture[(j+2) % puncture_len] && ((coded[j+2] > 0) != __builtin_parity(r & g3)))
+        if (puncture[(j+2) % puncture_len] && ((coded[j+2] > 0) != __builtin_parity(r & gens[2])))
             errors++;
     }
 
     return errors;
 }
 
-static int bit_errors_p1_fm(int8_t *coded, uint8_t *decoded)
+static int bit_errors_p_fm(int8_t *coded, uint8_t *decoded, const int len)
```
argilo: The various FM punctured codes are named for their coding rates, with this one called the "Rate 2/5 Convolutional Encoder". So maybe `2_5` would make more sense than `p`.
```suggestion
static int bit_errors_2_5_fm(int8_t *coded, uint8_t *decoded, const int len)
```

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
-    for (i = 0; i < P1_FRAME_LEN_ENCODED_FM; i++)
+    for (i = 0; i < N; i++)
     {
-        int partition = v[i % J];
-        int block = ((i / J) + (partition * 7)) % B;
-        int k = i / (J * B);
-        int row = (k * 11) % 32;
-        int column = (k * 11 + k / (32*9)) % C;
-        st->viterbi_p1[out++] = st->buffer_pm[(block * 32 + row) * 720 + partition * C + column];
+        const int partition = V[((i + 2 * (M / 4)) / M) % J];
```
argilo: If we care about supporting advanced service modes, we should probably pass in the length of `V` and use that instead of `J` here. From the spec:

> For standard service modes, either the number of interleaver partitions `J` or `length(v)` could be used as the modulus in the above partition index equation. However, since the two quantities can differ in some advanced service modes, only `length(v)` generally serves as the proper modulus for all standard and advanced service modes.

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
+        const int block = ((i / J) + (partition * 7)) % B;
```
argilo: This equation is different for the case when M = 2 or 4, so I suppose we should handle that case.

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
-        int k = ((i / J) % (PIDS_FRAME_LEN_ENCODED_FM / J)) + (P1_FRAME_LEN_ENCODED_FM / (J * B));
-        int row = (k * 11) % 32;
-        int column = (k * 11 + k / (32*9)) % C;
-        st->viterbi_pids[out++] = st->buffer_pm[(block * 32 + row) * 720 + partition * C + column];
+        const int partition = V[i % J];
+        const int block = i / b;
+        const int k = ((i / J) % (N / J)) + (I0 / (J * B));
+        const int row = (k * 11) % 32;
+        const int column = (k * 11 + k / (32*9)) % C;
+        viterbi[out++] = in[(block * 32 + row) * 720 + partition * C + column];
```
argilo:
```suggestion
        viterbi[out++] = in[(block * 32 + row) * (J * C) + partition * C + column];
```

src/decode.c
TheDaChicken: Didn't know a good name. That seems a bit better. "p" is misleading

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
-        int k = i / (J * B);
-        int row = (k * 11) % 32;
-        int column = (k * 11 + k / (32*9)) % C;
-        st->viterbi_p1[out++] = st->buffer_pm[(block * 32 + row) * 720 + partition * C + column];
+        const int partition = V[((i + 2 * (M / 4)) / M) % J];
+        const int block = ((i / J) + (partition * 7)) % B;
+        const int k = i / (J * B);
+        const int row = (k * 11) % 32;
+        const int column = (k * 11 + k / (32 * 9)) % C;
+        viterbi[out++] = in[(block * 32 + row) * 720 + partition * C + column];
```
argilo:
```suggestion
        viterbi[out++] = in[(block * 32 + row) * (J * C) + partition * C + column];
```

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
+
+    for (int i = bc * b; i < (bc+1) * b; i++)
     {
-        int partition = v[i % J];
-        int block = decode_get_block(st) - 1;
-        int k = ((i / J) % (PIDS_FRAME_LEN_ENCODED_FM / J)) + (P1_FRAME_LEN_ENCODED_FM / (J * B));
-        int row = (k * 11) % 32;
-        int column = (k * 11 + k / (32*9)) % C;
-        st->viterbi_pids[out++] = st->buffer_pm[(block * 32 + row) * 720 + partition * C + column];
+        const int partition = V[i % J];
```
argilo: As above, `J` should be `length(v)`.

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
+        const int block = i / b;
+        const int k = ((i / J) % (N / J)) + (I0 / (J * B));
```
argilo:
```suggestion
        const int k = ((i / J) % (b / J)) + (I0 / (J * B));
```

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
+        const int row = (k * 11) % 32;
+        const int column = (k * 11 + k / (32*9)) % C;
```
argilo: Spacing was changed for `inteleaver_i`, so may as well do it here too.
```suggestion
        const int column = (k * 11 + k / (32 * 9)) % C;
```

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
-        int column = (pti * 11) % C;
+        interleaver->i = 0;
+        memset(interleaver->pt, 0, sizeof(unsigned int) * 4);
+        interleaver->ready = 1;
+    }
+
+    unsigned int out = 0;
+    for (unsigned int i = 0; i < interleaver->frame_length * 2; i++)
+    {
+        const int partition = ((interleaver->i + 2 * (M / 4)) / M) % J;
```
argilo: This `J` is also technically `length(v)`, but I think we can leave it as is, since there aren't any modes that have `J != length(v)` when using Interleaver IV.

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
-    pids_frame_push(&st->pids, st->scrambler_pids);
+            viterbi[out++] = 0;
+    };
 }
 
-void decode_process_p3_p4(decode_t *st, interleaver_iv_t *interleaver, int8_t *viterbi, uint8_t *scrambler, unsigned int frame_len, logical_channel_t lc)
+void interleaver_iv(interleaver_iv_t* interleaver, int8_t* viterbi)
 {
-    const unsigned int J = (frame_len == P3_FRAME_LEN_FM) ? 4 : 2;
+    const unsigned int N = (interleaver->frame_length == P3_FRAME_LEN_MP3_MP11) ? 147456 : 73728;
```
argilo: These values were ordered like the parameter tables in the spec, so maybe `N` should go back down below `M`.

src/decode.c
```diff
@@ -268,96 +291,185 @@ static void descramble(uint8_t *buf, unsigned int length)
...
+    memcpy(st->buffer_t + (bc * BLKSZ * PARTITION_WIDTH_AM), sym_t, BLKSZ * PARTITION_WIDTH_AM);
+
+    decode_process_p1_p3_am(st, bc);
+}
+
+void decode_process_p1(decode_t *st)
+{
+    const int J = 20, B = 16, C = 36;
+    interleaver_i(st->buffer_pm, st->viterbi_p1,
+        J, B, C, 1, V_PM, P1_FRAME_LEN_ENCODED_FM);
```
argilo:
```suggestion
    const int J = 20, B = 16, C = 36, M = 1;
    interleaver_i(st->buffer_pm, st->viterbi_p1,
        J, B, C, M, V_PM, P1_FRAME_LEN_ENCODED_FM);
```

src/decode.c
```diff
@@ -369,42 +481,40 @@ void decode_process_pids_am(decode_t *st)
...
+    const int pids1_disabled = (st->input->sync.psmi == 1) && st->input->sync.rdbi;
     for (int i = 0; i < 10; i++) {
       for (int j = 0; j < 12; j++) {
         st->viterbi_pids[i*24 + pids_il_delay[j]] = pids1_disabled ? 0 : (il[i*12 + j] ? 1 : -1);
         st->viterbi_pids[i*24 + pids_iu_delay[j]] = iu[i*12 + j] ? 1 : -1;
       }
     }
 
-    nrsc5_conv_decode_e3(st->viterbi_pids, st->scrambler_pids, PIDS_FRAME_LEN);
+    nrsc5_conv_decode_e2(st->viterbi_pids, st->scrambler_pids, PIDS_FRAME_LEN);
```
argilo: PIDS uses the E3 convolutional decoder. If we merge E2 and E3 (since they share the same mother code, and differ only in puncturing) we can rename to `e2_e3` then.

src/decode.c
```diff
@@ -441,32 +551,24 @@ void decode_process_p1_p3_am(decode_t *st)
     }
 }
 
-void decode_set_block(decode_t *st, unsigned int bc)
+int decode_set_psmi(decode_t *st, const int mode, const unsigned int psmi)
```
argilo: Is it necessary to pass in `mode` here? It's in `st->radio->mode`, isn't it?

argilo: Do you anticipate some use for the return value?

TheDaChicken: Its to reduce the circular dependency of input containing decoder and decoder using input.

TheDaChicken: Ah your talking about radio struct. Similar situation i think.

argilo: Hmm, I guess decoder doesn't actually have a pointer to the parent radio, so it would have to go through input.

I see that elsewhere in the decoder we access the PSMI value with `st->input->sync.psmi`.

it feels a bit weird to have something that looks like a setter but doesn't actually store the value of psmi, just adjusts one other thing that depends on it.

To be honest, with the current design I like the old `decode_set_px1_length` function better. If we want to reorganize the structures to uncouple them, maybe it's best to leave that to a separate project.

argilo: On the other hand, I'm starting to see some logic in the change, since why should the sync stage have to know or care about PX1?

argilo: Here's an idea: Could we get rid of this function completely, and instead pass a length in with the call to `decode_push_px1` and `decode_push_px2`?

argilo: I expect then the `frame_length` struct field won't be needed at all.

src/decode.c
```diff
@@ -413,26 +525,26 @@ void decode_process_p1_p3_am(decode_t *st)
                 if (st->input->sync.psmi != SERVICE_MODE_MA3)
                 {
                     total_frame_length += P3_FRAME_LEN_ENCODED_MA1;
-                    nrsc5_conv_decode_e2(st->viterbi_p3_am, st->scrambler_p3_am, P3_FRAME_LEN_MA1);
-                    st->am_errors += bit_errors_p3_ma1(st->viterbi_p3_am, st->scrambler_p3_am);
+                    nrsc5_conv_decode_e1(st->viterbi_p3_am, st->scrambler_p3_am, P3_FRAME_LEN_MA1);
```
argilo: This looks wrong. MA1 uses the E2 convolutional encoder for P3.

Issue #453, opened 2025-09-22
Add clock synced to the HD signal?


pclov3r:
It might be nice if we had a feature to show a clock that was synced to HD radio signal time and time zone.

PR #504, opened 2026-04-25
ALFN Decode & Add to API

TheDaChicken:
ALFN (Absolute L1 Frame Number) is a number used to reference transmissions to absolute time. This field is required after `leap_seconds_offset` to calculate the current UTC timestamp from the station (#453 ). 

ALFN is sent through PIDS logical channel and different for AM and FM. AM was a pain in the butt. I was questioning my sanity. 

This PR only decodes such number and provides the number through the API. The number is then printed out using debug logging.

Calculation to resolve transmitted "time" from alfn is like so:
```const time_t utc = (65536 * (time_t)alfn / 44100) + 315964800 - (time_t)st->current_leap;``` This information is not printed due to tons of messed up stations.

Some reasons why to incorporate this silly number other than the terrible GPS timing:
- Relative time to determine what L1 frame you are at
- Allows someone else to investigate this time nonsense. 
- See what's being outputted like from gr-nrsc5 or a local station.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -613,8 +613,8 @@ struct nrsc5_event_t
         } local_time;
         struct
         {
-            unsigned int alfn;  /**< Absolute L1 Frame Number (ALFN) of the previous L1 Frame. */
-            int gps_locked;     /**< 1 if GPS is locked, otherwise 0. */
+            unsigned int alfn;  /**< Absolute L1 Frame Number (ALFN) of the most recent L1 Frame. */
+            int time_locked;    /**< 1 if time is locked to GPS, otherwise 0. */
         } alfn;
     };
 };
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -192,6 +192,7 @@ enum
     NRSC5_EVENT_IMPORTER_INFO,
     NRSC5_EVENT_LEAP_SECOND_OFFSET,
     NRSC5_EVENT_LOCAL_TIME,
+    NRSC5_EVENT_ALFN,
 };
 
 enum
@@ -401,6 +402,7 @@ struct nrsc5_event_t
  * - `NRSC5_EVENT_IMPORTER_INFO` : importer data, see `importer_info` member
  * - `NRSC5_EVENT_LEAP_SECOND_OFFSET` : leap second offset, see `leap_second_offset` member
  * - `NRSC5_EVENT_LOCAL_TIME` : local time data, see `local_time` member
+ * - `NRSC5_EVENT_ALFN` : ALFN data, see `alfn` member
  */
     unsigned int event;
     union
@@ -609,6 +611,11 @@ struct nrsc5_event_t
             int dst_local;     /**< 1 if DST is practiced locally, otherwise 0. */
             int dst_schedule;  /**< DST Schedule. 0 means Daylight Saving Time is not practiced. 1 means U.S./Canada schedule. 2 means EU schedule. */
         } local_time;
+        struct
+        {
+            unsigned int alfn;  /**< Absolute L1 Frame Number (ALFN) of the previous L1 Frame. */
+            int gps_locked;     /**< 1 if GPS is locked, otherwise 0. */
+        } alfn;
     };
 };
 /**
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -192,7 +192,7 @@ enum
     NRSC5_EVENT_IMPORTER_INFO,
     NRSC5_EVENT_LEAP_SECOND_OFFSET,
     NRSC5_EVENT_LOCAL_TIME,
-    NRSC5_EVENT_ALFN,
+    NRSC5_EVENT_L1_FRAME,
 };
 
 enum
@@ -402,7 +402,7 @@ struct nrsc5_event_t
  * - `NRSC5_EVENT_IMPORTER_INFO` : importer data, see `importer_info` member
  * - `NRSC5_EVENT_LEAP_SECOND_OFFSET` : leap second offset, see `leap_second_offset` member
  * - `NRSC5_EVENT_LOCAL_TIME` : local time data, see `local_time` member
- * - `NRSC5_EVENT_ALFN` : ALFN data, see `alfn` member
+ * - `NRSC5_EVENT_L1_FRAME` : L1 frame data, see `l1_frame` member
  */
     unsigned int event;
     union
@@ -614,8 +614,9 @@ struct nrsc5_event_t
         struct
         {
             unsigned int alfn;  /**< Absolute L1 Frame Number (ALFN) of the most recent L1 Frame. */
+            int alfn_known;     /**< 1 if ALFN is known, otherwise 0. */
             int time_locked;    /**< 1 if time is locked to GPS, otherwise 0. */
-        } alfn;
+        } l1_frame;
     };
 };
 /**
--- a/src/decode.c
+++ b/src/decode.c
@@ -468,10 +468,13 @@ void decode_process_pids(decode_t *st, const unsigned int bc)
 
     nrsc5_conv_decode_pids(st->viterbi_pids, st->scrambler_pids);
     descramble(st->scrambler_pids, PIDS_FRAME_LEN);
-    pids_frame_push(&st->pids, st->scrambler_pids);
+    pids_frame_push(&st->pids, st->scrambler_pids, bc);
+
+    if (bc == 15)
+        pids_complete_fm(&st->pids);
 }
 
-void decode_process_pids_am(decode_t *st, const uint8_t* sbit)
+void decode_process_pids_am(decode_t *st, const uint8_t* sbit, const unsigned int bc)
 {
     uint8_t il[120], iu[120];
 
@@ -501,7 +504,10 @@ void decode_process_pids_am(decode_t *st, const uint8_t* sbit)
 
     nrsc5_conv_decode_e2_e3(st->viterbi_pids, st->scrambler_pids, PIDS_FRAME_LEN);
     descramble(st->scrambler_pids, PIDS_FRAME_LEN);
-    pids_frame_push(&st->pids, st->scrambler_pids);
+    pids_frame_push(&st->pids, st->scrambler_pids, bc);
+
+    if (bc == 7)
+        pids_complete_am(&st->pids);
 }
 
 void decode_process_p1_p3_am(decode_t *st, const unsigned int bc)
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -1162,7 +1162,7 @@ void nrsc5_report_alfn(nrsc5_t* st, const unsigned int alfn, const int gps_locke
 
     evt.event = NRSC5_EVENT_ALFN;
     evt.alfn.alfn = alfn;
-    evt.alfn.gps_locked = gps_locked;
+    evt.alfn.time_locked = gps_locked;
 
     nrsc5_report(st, &evt);
 }
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -1156,6 +1156,17 @@ void nrsc5_report_local_time(nrsc5_t *st, const int tzo, const int dst_regional,
     nrsc5_report(st, &evt);
 }
 
+void nrsc5_report_alfn(nrsc5_t* st, const unsigned int alfn, const int gps_locked)
+{
+    nrsc5_event_t evt;
+
+    evt.event = NRSC5_EVENT_ALFN;
+    evt.alfn.alfn = alfn;
+    evt.alfn.gps_locked = gps_locked;
+
+    nrsc5_report(st, &evt);
+}
+
 void nrsc5_report_here_image(nrsc5_t *st, int image_type, int seq, int n1, int n2, unsigned int timestamp,
                              float latitude1, float longitude1, float latitude2, float longitude2,
                              const char *name, unsigned int size, const uint8_t *data)
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -1156,13 +1156,14 @@ void nrsc5_report_local_time(nrsc5_t *st, const int tzo, const int dst_regional,
     nrsc5_report(st, &evt);
 }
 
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

File: include/nrsc5.h [hunk 1]
```diff
@@ -609,6 +611,11 @@ struct nrsc5_event_t
             int dst_local;     /**< 1 if DST is practiced locally, otherwise 0. */
             int dst_schedule;  /**< DST Schedule. 0 means Daylight Saving Time is not practiced. 1 means U.S./Canada schedule. 2 means EU schedule. */
         } local_time;
+        struct
+        {
+            unsigned int alfn;  /**< Absolute L1 Frame Number (ALFN) of the previous L1 Frame. */
```
```suggestion
            unsigned int alfn;  /**< Absolute L1 Frame Number (ALFN) of the most recent L1 Frame. */
```

include/nrsc5.h
```diff
@@ -609,6 +611,11 @@ struct nrsc5_event_t
...
+            int gps_locked;     /**< 1 if GPS is locked, otherwise 0. */
```
argilo: The spec calls this field "Time Locked", so maybe we should use that.
```suggestion
            int time_locked;    /**< 1 if time is locked to GPS, otherwise 0. */
```

support/nrsc5.py
```diff
@@ -655,7 +663,8 @@ class _EventUnion(ctypes.Union):
         ("exciter_info", _ExciterInfo),
         ("importer_info", _ImporterInfo),
         ("leap_second_offset", _LeapSecondOffset),
-        ("local_time", _LocalTime)
+        ("local_time", _LocalTime),
+        ("alfn", _ALFN)
```
argilo: Maybe we should throw in a trailing comma so diffs are nicer in the future.
```suggestion
        ("alfn", _ALFN),
```

support/cli.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import argparse
+import datetime
```
argilo: This import is no longer used.

support/cli.py
```diff
@@ -398,6 +402,9 @@ def callback(self, evt_type, evt):
                           evt.utc_offset,
                           evt.dst_schedule,
                           "yes" if evt.dst_regional else "no", "yes" if evt.dst_local else "no")
+        elif evt_type == nrsc5.EventType.ALFN:
+            alfn = evt.alfn + 1
```
argilo: Since we're not using this for time calculations, there's probably no point in adding one anymore.

src/decode.c
```diff
@@ -468,10 +468,13 @@ void decode_process_pids(decode_t *st, const unsigned int bc)
 
     nrsc5_conv_decode_pids(st->viterbi_pids, st->scrambler_pids);
     descramble(st->scrambler_pids, PIDS_FRAME_LEN);
-    pids_frame_push(&st->pids, st->scrambler_pids);
+    pids_frame_push(&st->pids, st->scrambler_pids, bc);
+
+    if (bc == 15)
+        pids_complete_fm(&st->pids);
```
argilo, line 474: The decode stage having to tell the PIDS decoder when a frame is complete seems a bit messy. Do you think it might make sense to instead pass in the mode (FM/AM) as an argument to `pids_frame_push` so that the PIDS decoder can take care of this check?

include/nrsc5.h
TheDaChicken: Isn't describing this as "most recent" misleading? Isn't this the previous L1 frame number?

argilo: I don't understand what was meant by "previous". After we decode an L1 frame, we report its L1 frame number.

src/decode.c
TheDaChicken: I could do that. That means I will need to check for bc == 15 or bc == 6 in pids. I was trying to reduce that.

TheDaChicken: Providing the mode as an argument to pids_frame_push does mean I have to do this:
```
    if (mode == NRSC5_MODE_FM && bc == 15)
        pids_complete_fm(st);
    if (mode == NRSC5_MODE_AM && bc == 8)
        pids_complete_am(st);
 ```

Comment (argilo) [reply src/decode.c:474]: I guess there's an extra mode check if you put it inside, but it does seem better from an encapsulation perspective.

File: include/nrsc5.h (hunk 1 above)
Comment (TheDaChicken) [reply include/nrsc5.h]: What I mean is that once we receive the full ALFN number, the L1 frame is finished. Most recent misses that detail.

File: src/decode.c (hunk 6 above)
Comment (TheDaChicken) [reply src/decode.c:474]: Yeah.. I guess so. Whatever makes the code less messy especially long term.

File: include/nrsc5.h (hunk 1 above)
Comment (argilo) [reply include/nrsc5.h]: How about "most recently received" then? I think that would imply that the whole frame has been received. To me "previous" sounds like it would be the one before that.

Comment (TheDaChicken) [reply include/nrsc5.h]: I can see that "previous" sounds like its the one before. 

I think most recently received is better but wordy. Maybe this is better? 
```
/**< Absolute L1 Frame Number (ALFN) of the most recent L1 Frame. Refers to the start time of the frame. */
```
This may sound a bit misleading still?

Comment (argilo) [reply include/nrsc5.h]: > Refers to the start time of the frame.

I wonder if this is a detail we need to get into. While the standard does envision that frames should be transmitted such that their start time occurs at the ALFN number times 65536/44100 seconds after the start of the GPS epoch, in practice ALFN seems to be just a sequential number that is sometimes in the rough vicinity of GPS time divided by 65536/44100.

File: src/decode.c (hunk 6 above)
Comment (TheDaChicken) [reply src/decode.c:474]: Would this cause issues with Advanced services modes where there's many PIDS frames?

File: include/nrsc5.h (hunk 1 above)
Comment (TheDaChicken) [reply include/nrsc5.h]: Yeah, I think that's a valid point for not including that detail. The detail is basically nonsensical at this point. I'll just commit your suggested "most recent" change :)

File: src/decode.c (hunk 6 above)
Comment (argilo) [reply src/decode.c:474]: Good question. To be honest, I haven't read up on them to know exactly how they work. But we should probably design things in such a way that we can deal with the increased PIDS bandwidth and presence of LLDS PDUs in the future.

It could be that the current design where the decode layer informs the PIDS layer about the end of a frame works better.

File: support/cli.py [hunk 7]
```diff
@@ -398,6 +401,9 @@ def callback(self, evt_type, evt):
...
+            logging.debug("ALFN %u, GPS locked? %s", alfn,"yes" if evt.time_locked else "no")
```
argilo: Perhaps we could be a bit more explicit in describing `time_locked` (here and in the C client):
```suggestion
            logging.debug("ALFN %u, time locked to GPS? %s", alfn, "yes" if evt.time_locked else "no")
```

src/nrsc5.c
```diff
@@ -1156,6 +1156,17 @@ void nrsc5_report_local_time(nrsc5_t *st, const int tzo, const int dst_regional,
     nrsc5_report(st, &evt);
 }
 
+void nrsc5_report_alfn(nrsc5_t* st, const unsigned int alfn, const int gps_locked)
```
argilo: Rename the argument too?
```suggestion
void nrsc5_report_alfn(nrsc5_t* st, const unsigned int alfn, const int time_locked)
```

src/pids.c
```diff
@@ -1018,6 +1018,16 @@ static void sis_decode(pids_t *st, const uint8_t *bits)
         }
     }
 
+    const uint8_t time_locked = bits[64];
+    const uint8_t adv_alfn = bits[65] << 1 | bits[66];
+
+    alfn_t* alfn = &st->alfn[st->alfn_frame];
+
+    alfn->value |= (uint32_t)adv_alfn << (2 * bc);
+    alfn->received |= (1 << bc);
```
argilo: I wonder if we should do something like this, so that we only take the ALFN bits from the first valid PIDS PDU in each L1 block? I think ORing together multiple values would increase the probability of error in low-SNR signals where corrupt PIDS PDUs might occasionally make it past the CRC check.
```suggestion
    if ((alfn->received & (1 << bc)) == 0)
    {
        alfn->value |= (uint32_t)adv_alfn << (2 * bc);
        alfn->received |= (1 << bc);
    }
```

src/pids.c
```diff
@@ -1098,5 +1192,14 @@ void pids_init(pids_t *st, input_t *input)
 
     reset_alert(st);
 
+    st->alfn_frame = 0;
+    st->alfn_frame = 0;
```
argilo: This line is duplicated.

src/pids.c
TheDaChicken: Ah, I think that's a good idea

src/pids.c
```diff
@@ -1029,7 +1039,91 @@ static void sis_decode(pids_t *st, const uint8_t *bits)
...
+            const int lower[ALFN_AM_FRAMES] = {
+                (int)(st->alfn[0].value & 0x3),
+                (int)(st->alfn[1].value & 0x3),
+                (int)(st->alfn[2].value & 0x3),
+                (int)(st->alfn[3].value & 0x3),
+            };
+            int finished = 1;
+            for (int i = 0; i < ALFN_AM_FRAMES; i++)
+            {
+                if ((st->alfn[i].received & 0x1) == 0)
```
argilo: Why are we only checking one bit of `received` here?

argilo: Oh, I guess it's because we only need the two least significant bits of the ALFN to do the `find_mod4` thing. Never mind.

TheDaChicken: Yeahh. Its probably self explanatory why this is useful. We can still decode some ALFNs even during low SNR situations where only one upper and one lower has been fully received.

src/pids.c
```diff
@@ -1029,7 +1039,91 @@ static void sis_decode(pids_t *st, const uint8_t *bits)
...
+
+    const alfn_t* upper_frame = &st->alfn[st->alfn_upper_idx];
+    const alfn_t* lower_frame = &st->alfn[lower_idx];
+
+    if (upper_frame->received != 0xFF ||
+        lower_frame->received != 0xFF)
+        return;
+
+    const uint16_t upper = upper_frame->value & 0xFFFF;
+    uint16_t lower = lower_frame->value & 0xFFFF;
```
argilo: Are the `& 0xFFFF`'s necessary here? I expect there's no way `value` could have higher bits set.

TheDaChicken: Yeah, it's technically unnecessary now. I wanted to make `upper_frame->value ` clear that the value is 16 bits for AM when the type is 32.

TheDaChicken: I can remove that

src/pids.c
```diff
@@ -1029,7 +1039,91 @@ static void sis_decode(pids_t *st, const uint8_t *bits)
...
+        }
+    }
+    return -1;
+}
+
+void pids_complete_am(pids_t* st)
+{
+    if (st->alfn[st->alfn_frame].received == 0xFF)
+    {
+        if (st->alfn_frame == 3 && st->alfn_upper_idx == -1)
```
argilo: Is there a reason for this section of code to be inside the `received == 0xFF` check, given that the `find_mod4` stuff only cares about the two least significant bits?

TheDaChicken: Omg, you are right. This is from old iterations of the code. You are correct that theres no reason for this to be inside received == 0xFF  anymore.

argilo: Also, is there a reason we're only attempting to find the upper index when `st->alfn_frame == 3`?

TheDaChicken: Well, there's no reason to try to find ALFN upper if 3 frames haven't been received yet. Technically this could be considered unnecessarily when 3 frames have received already.

argilo: Lastly, the `st->alfn_upper_idx == -1` check worries me slightly, since it means that if we make the wrong decision, the decoder is poisoned forever. I suppose the likelihood of that is low though, since at least two of the four SIS PDUs involved would need to be corrupted for that to happen, and the chance of a single SIS PDU passing the CRC12 check is 1/4096.

TheDaChicken: Yeah, I can understand that. An idea is to reset alfn_upper_idx if an upper frame has been received incomplete or I could just repeatedly look for the upper frame

argilo: > Well, there's no reason to try to find ALFN upper if 3 frames haven't been received yet.

Sure, but what if the SNR is low, and PDUs are getting dropped due to invalid CRCs? We'll only check once every four frames which PDUs from the four most recent frames were received, when we'd have a better chance by considering any four consecutive frames.

argilo: This code only runs once per frame, so I'm not at all concerned about efficiency in terms of CPU utilization.

TheDaChicken: > We'll only check once every four frames which PDUs

Yeah, that's true. Possibly removing this change would improve decoding ALFN on AM a lot.

argilo: For that reason, it may be sensible to simplify a bit and get rid of some of the `if`s.

argilo:
One last thing that may be worth consideration:

If we fail to decode the ALFN value for a given frame, but we haven't lost sync, we could consider incrementing the most recent ALFN and returning that, since we know that ALFNs are sequential.

I'm not convinced this in necessarily worth doing, but it could be that clients would prefer to see an ALFN event for every L1 frame instead of having gaps.

If we went this route, I suppose it would also bring up the question of whether we should report ALFN events for L1 frames received before the first valid ALFN is decoded (with some sort of dummy value to indicate that the ALFN is unknown).

TheDaChicken:
I was thinking something similar but for something else. 

I think I see a point there from what your saying. Since ALFN can be used to identify the frame, it wouldn't be super useful if sometimes the ALFN isn't reported

TheDaChicken:
I can kind of see that being able to identify a frame may be useful for debugging

TheDaChicken:
For record purposes I do want to give an update,

Currently there's now an `l1_frame` event which includes fields such if time is locked to GPS, if alfn is known and the actual alfn number if known.

This seems like the best direction to go towards. Any other information related to L1 frames can be added to this event. The difficult part is predicting the use of this field.

PR #513, opened 2026-06-21
Refactor to use float input and native float pipeline

pclov3r:
This is a draft PR with the help of AI to draft some ideas on how move the entire nrsc5 pipeline to processing floats instead of integers. This makes it significantly easier to support other SDR applications, as most work with or convert to floats instead of using integers directly. 

In this PR:                       
                                                                                                                                       

- A new  nrsc5_pipe_samples_cf32  API call to pipe in floats was added.                                                              
- To maintain support for  cu8  and  cs16  sample input, we convert both to float within  input.c .                                  
- Replaced the old integer FIR decimator with a new float-based one ( firdecim_cf32 ). Manual intrinsics were removed, and we rely on compiler auto-vectorization now. Verified the GCC assembly output and it appears auto-vectorization is working correctly on both ARM and x86. This could be replaced with a resampler instead to avoid issues but that seems like it would increase CPU usage for no good reason.                                                                                                    
- Tweaked the  fread  chunking in  main.c  to prevent crashes if someone pipes in a truncated or corrupted file.

Since this is a pretty major change please review carefully for issues and suggest refinements. 

Thanks!


pclov3r:
Closing in light of #515

PR #515, opened 2026-06-23
Adds Float Input

### LINKED ISSUE #445: Float input?
I came across a problem using the API. My HF+ Discovery uses floats natively. If I wanted to resample, I have to then downscale to int16s or uint8 to get it passed to libnrsc5. 

While its possible to downscale, its annoying. Most libraries use floats as the datatype for complex numbers like GNU Radio, Liquiddsp. It would make an NRSC5 Decoder block much more easier to implement into GNU Radio if libnrsc5 accepted floats as input. 

This seems to not be that difficult as libnrsc5 already converts to floats internally:

https://github.com/theori-io/nrsc5/blob/310fa3792c5ea3436cf6d37590a93035ba86b566/src/acquire.c#L160-L161
https://github.com/theori-io/nrsc5/blob/310fa3792c5ea3436cf6d37590a93035ba86b566/src/defines.h#L97-L100

From my understanding, acquire.c also looks to contains an unnecessary extra buffer to hold int16s and converts back to floats twice. 

This is something I could probably make a PR for.

TheDaChicken:
Fixes #445
Adapted from #513 

Like pclover said, this significantly makes using other SDR libraries with the nrsc5 API a lot easier. For example, gnuradio, liquid-dsp. 

This switches out all code for firdecim from int16s to floats including the neon code. Whether or not neon improves performance needs to be investigated. 

Now piping cu8 converts straight to floats. I saw a 1.5 times performance improvement with this.  

Due to this, the code for converting int16s to floats is now in `input_push_cs16` instead of acquire. 

This also adds a new constant `NRSC5_SAMPLE_RATE_NATIVE_FM` to be used for both cs16 and cf32. This depreciates both `NRSC5_SAMPLE_RATE_CS16_FM` and `NRSC5_SAMPLE_RATE_CS16_AM`.


README.md
```diff
@@ -96,37 +96,37 @@ Replace `64` with `32` if you want a 32-bit build. Once the build is complete, c
...
+                                           (0, 1, 2, or 3)
+    -g gain                              gain
+                                           (example: 49.6)
+                                           (automatic gain selection if not specified)
+    -d device-index                      rtl-sdr device
+    -p ppm-error                         rtl-sdr ppm error
+    -H rtltcp-host                       rtl_tcp host with optional port
+                                           (example: localhost:1234)
+    -r iq-input                          read IQ samples from input file
+    --iq-input-format {cu8,cs16,cf32}      IQ input format (complex unsigned 8-bit @ 1488375 SPS or
```
pclov3r: complex float 32-bit @ 744188 SPS (FM) / 46512 SPS (AM). is missing here in the readme.

TheDaChicken: Ah! Got it. Thanks.

src/nrsc5.c
```diff
@@ -649,6 +649,31 @@ int nrsc5_pipe_samples_cs16(nrsc5_t *st, const int16_t *samples, unsigned int le
...
+        st->leftover_f32_num = 0;
+    }
+
+    sample_groups = length / 2;
+    input_push_f32(&st->input, samples, sample_groups * 2);
+    samples += (sample_groups * 2);
+    length -= (sample_groups * 2);
+
+    if (length == 1)
+        st->leftover_s16[st->leftover_s16_num++] = samples[0];
```
pclov3r: This should be f32 not s16?

include/nrsc5.h
```diff
@@ -50,10 +50,12 @@
...
 
-#define NRSC5_SAMPLE_RATE_CU8      1488375     /**< Sample rate at which nrsc5_pipe_samples_cu8() expects samples (FM or AM) */
-#define NRSC5_SAMPLE_RATE_CS16_FM  744187.5    /**< Sample rate at which nrsc5_pipe_samples_cs16() expects samples (FM only) */
-#define NRSC5_SAMPLE_RATE_CS16_AM  46511.71875 /**< Sample rate at which nrsc5_pipe_samples_cs16() expects samples (AM only) */
-#define NRSC5_SAMPLE_RATE_AUDIO    44100       /**< Sample rate of outgoing audio */
+#define NRSC5_SAMPLE_RATE_CU8      1488375            /**< Sample rate at which nrsc5_pipe_samples_cu8() expects samples (FM or AM) */
+#define NRSC5_SAMPLE_RATE_NATIVE_FM  744187.5         /**< Sample rate at which nrsc5_pipe_samples_cs16() & nrsc5_pipe_samples_cf32() expects samples (FM only) */
+#define NRSC5_SAMPLE_RATE_NATIVE_AM  46511.71875      /**< Sample rate at which nrsc5_pipe_samples_cs16() & nrsc5_pipe_samples_cf32() expects samples (AM only) */
+#define NRSC5_SAMPLE_RATE_CS16_FM  NRSC5_SAMPLE_RATE_NATIVE_FM    /**< DEPRECATED: use `NRSC5_SAMPLE_RATE_NATIVE_FM` instead */
+#define NRSC5_SAMPLE_RATE_CS16_AM  NRSC5_SAMPLE_RATE_NATIVE_AM      /**< DEPRECATED: use `NRSC5_SAMPLE_RATE_NATIVE_Am` instead */
```
```suggestion
#define NRSC5_SAMPLE_RATE_CS16_AM  NRSC5_SAMPLE_RATE_NATIVE_AM      /**< DEPRECATED: use `NRSC5_SAMPLE_RATE_NATIVE_AM` instead */
```

src/nrsc5.c
```diff
@@ -649,6 +649,31 @@ int nrsc5_pipe_samples_cs16(nrsc5_t *st, const int16_t *samples, unsigned int le
...
+
+    if (st->leftover_f32_num == 1 && length > 0)
+    {
+        st->leftover_f32[st->leftover_f32_num++] = samples[0];
+        samples++;
+        length--;
+
+        input_push_f32(&st->input, st->leftover_f32, 2);
+        st->leftover_f32_num = 0;
+    }
```
argilo: In hindsight, I think I may have gone overboard in #328 by allowing the I and Q parts of a sample to be broken up. It brings in the extra complexity of having to track the "leftovers." I was actually considering removing that from `nrsc5_pipe_samples_cs16`, so I wonder whether it might sense to avoid adding it to `nrsc5_pipe_samples_cf32` in the first place and just require `length` to be even.

argilo: I could be convinced otherwise, if the flexibility is of some use to API clients.

TheDaChicken: I don't see this super useful either since these are only meant to accept complex numbers. Reading a file maybe is helpful if you are reading based on bytes?

argilo: > Reading a file maybe is helpful if you are reading based on bytes?

I could see that maybe being the case for CU8, but for CS16 and CF32 the API client would already have to ensure it's dealing with 2 or 4 bytes at a time, since those are the sizes of `int16_t` and `float`.

TheDaChicken: Well, originally I complained because CU8 required a multiple of 4 complex samples. For CS16, I didn't have this issue. I'll remove this for cf32. I would rather not include more of this code unless needed.

src/main.c
```diff
@@ -930,11 +936,12 @@ static int parse_args(state_t *st, int argc, char *argv[])
 
     if (st->input_name && (st->iq_input_format == IQ_FORMAT_NONE))
     {
-        const char *suffix = ".cs16";
-        size_t suffix_len = strlen(suffix);
-        size_t len = strlen(st->input_name);
-        if ((len >= suffix_len) && (strcmp(st->input_name + len - suffix_len, suffix) == 0))
+        const size_t len = strlen(st->input_name);
```
argilo: It might be cleaner to move the `strlen` into `ends_with` and get rid of the `len` argument.

support/nrsc5.py
```diff
@@ -1062,3 +1064,10 @@ def pipe_samples_cs16(self, samples):
         result = NRSC5.libnrsc5.nrsc5_pipe_samples_cs16(self.radio, samples, len(samples) // 2)
         if result != 0:
             raise NRSC5Error("Failed to pipe samples.")
+
+    def pipe_samples_cf32(self, samples):
+        if len(samples) % 8 != 0:
+            raise NRSC5Error("len(samples) must be a multiple of 8.")
+        result = NRSC5.libnrsc5.nrsc5_pipe_samples_cf32(self.radio, samples, len(samples) // 8)
```
argilo: This should still be `// 4`, shouldn't it? Since it counts the number of floats, not the number of I/Q samples.

```suggestion
        result = NRSC5.libnrsc5.nrsc5_pipe_samples_cf32(self.radio, samples, len(samples) // 4)
```

src/input.h
```diff
@@ -41,3 +41,4 @@ void input_free(input_t *st);
 void input_set_sync_state(input_t *st, unsigned int new_state);
 void input_push_cu8(input_t *st, const uint8_t *buf, uint32_t len);
 void input_push_cs16(input_t *st, const int16_t *buf, uint32_t len);
+void input_push_f32(input_t *st, const float *buf, uint32_t len);
```
argilo: `cf32` would probably be a better name, to reflect the fact that these are complex samples.
```suggestion
void input_push_cf32(input_t *st, const float *buf, uint32_t len);
```

argilo:
I did a first pass over the code, and the changes look good apart from a few things I commented on.

I also ran this against my test suite, and got roughly the same number of audio packets out. So the demodulation still appears to be correct.

I still want to do some further testing (verifying that the NEON code runs correctly, testing performance, and running with ASAN) before giving this the final thumbs up.

argilo:
A couple more changes are needed to make the new API function available publicly:
```diff
diff --git a/src/libnrsc5.map b/src/libnrsc5.map
index b118200..62a6d51 100644
--- a/src/libnrsc5.map
+++ b/src/libnrsc5.map
@@ -23,6 +23,7 @@ LIBNRSC5_1.0 {
         nrsc5_set_callback;
         nrsc5_pipe_samples_cu8;
         nrsc5_pipe_samples_cs16;
+        nrsc5_pipe_samples_cf32;
 
     local:
         *;
diff --git a/src/libnrsc5.sym b/src/libnrsc5.sym
index a241247..71cfc5d 100644
--- a/src/libnrsc5.sym
+++ b/src/libnrsc5.sym
@@ -21,3 +21,4 @@ _nrsc5_set_auto_gain
 _nrsc5_set_callback
 _nrsc5_pipe_samples_cu8
 _nrsc5_pipe_samples_cs16
+_nrsc5_pipe_samples_cf32
```

support/nrsc5.py
TheDaChicken: Yep, I think you are correct. That was me going too fast. The only thing that changed here was the multiple requirement.

src/main.c
TheDaChicken: Yeah. I wanted to "save" performance because it's a redundant operation. I don't think it saves anything. Compiler can deal with it.

argilo: And it's an operation that's done once at startup, so I don't think performance is a concern.

PR #517, opened 2026-06-26
Fetch libao from GitHub

argilo:
gitlab.xiph.org has been unreliable of late, causing Windows builds to fail.


pclov3r:
I wonder if it may make more sense to pull the code to a folder like external to avoid issues like this down the road? 

Given it targets a specific commit/release anyhow it might make sense to vendor the code?

argilo:
Nah, I'd rather not bloat our repo with code from dependencies. Fetching from GitHub seems fine to me, since nrsc5 itself is also hosted there.

PR #519, opened 2026-06-26
Remove NEON dot product functions

### LINKED ISSUE #516: Shoud nrsc5 contain neon code?
While looking through #513 , I wondered if the neon code for dotpod is worth keeping.

Currently to use the neon code, you must be using a 32 bit arm processor. Editing the CMakeList to enable the neon code for my 64 bit arm processor showed slower performance. Cpu time being spent on `halfband_fs32_execute` went from 7% to 9% via my #515 

Comparing current master branch, `halfband_q15_execute` cpu time went from 12.20 % to 14.79%

argilo:
Fixes #516.

As noted in #516, the NEON versions of `dotprod_32` and `dotprod_halfband_4` are slower than their generic counterparts, so let's remove them for now.

The removal of these functions also means that filter taps no longer need to be doubled, so I have adjusted the generic implementations accordingly. I had to re-insert the missing zero taps into `decim_taps` because `dotprod_halfband_4` was depending on the taps getting shifted into the right positions by the tap doubling.


argilo:
My test script shows identical output before & after this change, which is a good sign that I haven't broken anything.

ka9q:
Seems surprising that neon would be slower than scalar instructions. Is this true on all ARM/RPi models?

argilo:
Both of the functions involve small numbers of calculations (for example, the generic `dotprod_halfband_4` performs only 4 multiplies and 9 adds). I expect that the overhead of loading the data into NEON registers was enough to make the operation slower that the scalar version.

argilo:
The NEON `dotprod_32` did perform better than the generic version prior to the switch from 16-bit signed arithmetic to floating-point arithmetic in #515, because the NEON version could process 8 signed 16-bit integers at once, vs. 4 floating-point values now.

With some further optimization, it might be possible to make the new NEON version faster. But for now I like the simplicity of not having platform-specific code, and the performance of `dotprod_32` is not particularly important since it's only used for a short time during coarse OFDM synchronization.

argilo:
The impetus for optimizing `dotprod_32` was that nrsc5 originally had only coarse (pre-FFT) synchronization, which ran all the time and required an FIR filter.  Later, fine (post-FFT) synchronization was added, which eliminates the need for FIR filtering.

The `dotprod_halfband_4` function used for input decimation was added later, and I may have copied the idea over from `dotprod_32` without actually verifying whether NEON performance was better on the smaller input size.

ka9q:
Thanks. I'm interested in dot product speeds because there's usually a crossover point  in speed vs # taps between fast convolution with the FFT and a vectorized multiply-add FIR, and it hadn't really occurred to me that scalar might beat both for really short filters. Probably worth measuring on each platform if it gets executed at all often.

Issue #446, opened 2025-08-07
No way to easily reset nrsc5 session


TheDaChicken:
I found another API weird design. Currently in the API, there is no way to reset a nrsc5 session without providing a new frequency. `nrsc5_set_frequency` being the only function that can reset the state of libnrsc5.

When reading IQ from external sources like a file, I may want to reuse already existing nrsc5 session but I may not know what the frequency is.

argilo:
Resolved in #523.

PR #523, opened 2026-08-17
`nrsc5_reset` function

TheDaChicken:
#446 

Adds a `nrsc5_reset` function that resets the current nrsc5 session. Pretty simple and self explanatory I think. I also made ``nrsc5_reset`` reset the current auto gain. That might be helpful depending on the use case.

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -543,12 +543,20 @@ int nrsc5_set_frequency(nrsc5_t *st, float freq)
     if (st->rtltcp && rtltcp_set_center_freq(st->rtltcp, freq) != 0)
         return 1;
 
-    if (st->auto_gain)
-        st->gain = -1;
-    input_reset(&st->input);
-    output_reset(&st->output);
+    nrsc5_reset(st);
 
     st->freq = freq;
+    return 0;
+}
+
+NRSC5_API int nrsc5_reset(nrsc5_t *st) {
+    if (!st->stopped)
+        return 1;
+    if (st->auto_gain)
+        st->gain = -1;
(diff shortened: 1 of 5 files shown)


src/nrsc5.c
```diff
@@ -543,12 +543,20 @@ int nrsc5_set_frequency(nrsc5_t *st, float freq)
...
 
+    nrsc5_reset(st);
+
+    st->freq = freq;
+    return 0;
+}
+
+NRSC5_API int nrsc5_reset(nrsc5_t *st) {
+    if (!st->stopped)
+        return 1;
```
argilo, line 554: What is the logic behind this check?

argilo:
There are a few other places where `nrsc5_reset` needs to be added:

* libnrsc5.map
* libnrsc5.sym
* nrsc5.py

src/nrsc5.c
TheDaChicken: Its to make sure that rtlsdr sessions dont reset without stopping the session.

argilo: OK, but what goes wrong if the session is not stopped?

TheDaChicken: Thread-safety. This technically already doesn't make any sense since `nrsc5_set_frequency` also checks `stopped` without a mutex

argilo: Ah, I hadn't noticed that `nrsc5_set_frequency` already has the same check. I guess it makes sense for `nrsc5_reset` to have the same check then.

Thread safety would best be left to another issue / PR.

PR #524, opened 2026-08-18
Commercial tags to API.

TheDaChicken:
Continuation of #372 

Commercial tags are currently only printed out when libnrsc5 is compiled with debug logging. This adds commercial tags to the API and prints commercial tags in the CLIs. 

Comment tags were also completely broken if the encoding was UCS-2. `buf` and `len` were used instead of the proper `frame_len` and `data`

Example from CFMI:
```
22:11:09 Audio service 0: public, type: None, codec: 0, blend: 2, gain: 0 dB, delay: 96, latency: 8
22:11:09 Title: Wish You Were Here
22:11:09 Artist: Pink Floyd
22:11:09 Genre: ROCK
22:11:09 Commercial: price=USD000.00 until=2027-06-27 url="-" seller="-" desc="-" received_as=0
22:11:09 Audio service 2: public, type: None, codec: 0, blend: 0, gain: 0 dB, delay: 0, latency: 8
22:11:09 Audio service 1: public, type: None, codec: 0, blend: 0, gain: 0 dB, delay: 0, latency: 8
22:11:09 MER: 9.6 dB (lower), 11.4 dB (upper)
22:11:09 BER: 0.000358, avg: 0.000367, min: 0.000358, max: 0.000375
22:11:09 Title: Wish You Were Here
22:11:09 Artist: Pink Floyd
22:11:09 Genre: ROCK
22:11:09 Commercial: price=USD000.00 until=2027-06-27 url="-" seller="-" desc="-" received_as=0
```

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -455,12 +455,12 @@ struct nrsc5_event_t
             nrsc5_id3_comment_t *comments;
             struct
             {
-                char* price;
-                char valid_until[8];
-                char* contact_url;
-                char* seller;
-                char* description;
-                uint8_t received_as;
+                char* price;         /** price of the product */
+                char valid_until[8]; /** describes for how long the price is valid in YYYYMMDD format */
+                char* contact_url;   /** URL for contacting the seller */
+                char* seller;        /** Name of the seller */
+                char* description;   /** Short description of the product */
+                uint8_t received_as; /** How audio was delivered when bought */
             } commercial;
         } id3;
         struct {
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -453,6 +453,15 @@ struct nrsc5_event_t
                 int lot;
             } xhdr;
             nrsc5_id3_comment_t *comments;
+            struct
+            {
+                char* price;
+                char valid_until[8];
+                char* contact_url;
+                char* seller;
+                char* description;
+                uint8_t received_as;
+            } commercial;
         } id3;
         struct {
             uint16_t port;  /**< DEPRECATED: Use `component->data.port` instead */
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -455,12 +455,12 @@ struct nrsc5_event_t
             nrsc5_id3_comment_t *comments;
             struct
             {
-                char* price;            /** price of the product */
-                char* contact_url;      /** URL for contacting the seller */
-                char* seller;           /** Name of the seller */
-                char* description;      /** Short description of the product */
+                char *price;            /** price of the product */
+                char *contact_url;      /** URL for contacting the seller */
+                char *seller;           /** Name of the seller */
+                char *description;      /** Short description of the product */
                 uint8_t received_as;    /** How audio is delivered when bought */
-                struct tm* valid_until; /** Date when the price expires */
+                struct tm *valid_until; /** Date when the price expires. Note that only the `tm_year`, `tm_mon`, and `tm_mday` fields are set. */
             } commercial;
         } id3;
         struct {
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -455,12 +455,12 @@ struct nrsc5_event_t
             nrsc5_id3_comment_t *comments;
             struct
             {
-                char* price;         /** price of the product */
-                char valid_until[8]; /** describes for how long the price is valid in YYYYMMDD format */
-                char* contact_url;   /** URL for contacting the seller */
-                char* seller;        /** Name of the seller */
-                char* description;   /** Short description of the product */
-                uint8_t received_as; /** How audio was delivered when bought */
+                char* price;            /** price of the product */
+                char* contact_url;      /** URL for contacting the seller */
+                char* seller;           /** Name of the seller */
+                char* description;      /** Short description of the product */
+                uint8_t received_as;    /** How audio is delivered when bought */
+                struct tm* valid_until; /** Date when the price expires */
             } commercial;
         } id3;
         struct {
--- a/src/output.c
+++ b/src/output.c
@@ -274,7 +274,7 @@ static char *id3_text(uint8_t *buf, unsigned int frame_len)
         return id3_encode_utf8(0, NULL, 0);
 }
 
-static uint8_t* strlen_enc(const int enc, uint8_t *buf, const unsigned int len)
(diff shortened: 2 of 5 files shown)


include/nrsc5.h
```diff
@@ -453,6 +453,15 @@ struct nrsc5_event_t
                 int lot;
             } xhdr;
             nrsc5_id3_comment_t *comments;
+            struct
+            {
+                char* price;         /** price of the product */
+                char valid_until[8]; /** describes for how long the price is valid in YYYYMMDD format */
```
argilo: `char[8]` seems like an annoying format for API clients to deal with. Do you think it might be worth returning `int` fields for year, month, and day?

include/nrsc5.h
```diff
@@ -453,6 +453,15 @@ struct nrsc5_event_t
...
+                char* contact_url;   /** URL for contacting the seller */
+                char* seller;        /** Name of the seller */
+                char* description;   /** Short description of the product */
+                uint8_t received_as; /** How audio was delivered when bought */
```
```suggestion
                uint8_t received_as; /** How audio is delivered when bought */
```

argilo: Do you think we should go to the trouble of making an enumeration for the `received_as` field? Or maybe it would be enough to list the values in the documentation.

TheDaChicken: I think having an enum makes sense if we know that there are values other than zero being used. I don't mind adding them in an enum anyways just in case.

include/nrsc5.h
TheDaChicken: That sounds a lot better. Can we assume that the field will always be valid numbers?

argilo: It would probably make sense to validate and log an error (on the API side) if the value is not numeric. I'd say that situation is more likely to come up due to a bug in libnrsc5 than a station doing something wonky, but you never know.

TheDaChicken: Does having this information in a `struct tm` field make sense? That keeps everything consistent

argilo: I thought about that, but I worried that the extra fields might cause confusion. I suppose that could be addressed in the documentation, though.

argilo: I do like the idea of keeping the API consistent.

TheDaChicken: I think it works out. It's only the date and no time.

TheDaChicken:
I had to move `_TimeStruct` above ID3 because python requires a definition before usage. I was struggling and couldn't find any function in the std for converting string to int with bounds

support/nrsc5.py
```diff
@@ -754,9 +763,17 @@ def _callback_wrapper(self, c_evt):
...
                 comments.append(Comment(self._decode(c.lang), self._decode(c.short_content_desc), self._decode(c.full_text)))
                 comment_ptr = c.next
+            commercial = None
+            if id3.commercial.price:
+                commercial = Commercial(self._decode(id3.commercial.price),
+                                        self._decode(id3.commercial.contact_url),
+                                        self._decode(id3.commercial.seller),
+                                        self._decode(id3.commercial.description),
+                                        id3.commercial.received_as,
+                                        self._timestruct_to_datetime(id3.commercial.valid_until))
```
argilo: Python has a date-only `date` object. Do you suppose we should use that instead of `datetime`?

src/output.c
```diff
@@ -336,38 +376,59 @@ static void output_id3(output_t *st, unsigned int program, uint8_t *buf, unsigne
...
+                pos = delim[i] + (delim_enc == 1 ? 2 : 1);
                 if (i == 0)
+                {
+                    until.tm_year = parse_digits((char *) delim[0] + 1, 4) - 1900;
+                    until.tm_mon = parse_digits((char *) delim[0] + 5, 2) - 1;
+                    until.tm_mday = parse_digits((char *) delim[0] + 7, 2);
+
+                    if (until.tm_mday < 0 || until.tm_mon < 0 || until.tm_year < 0)
+                    {
+                        log_debug("Fail to parse valid_until on COMR tag.");
```
argilo: We should probably make this at least `warn` level since it could be indicative of a bug in libnrsc5.

support/nrsc5.py
TheDaChicken: Not a bad idea. Also while trying this I completely had forgotten that `_timestruct_to_datetime` constructs datetime with utc. There's no definition of timezone here.

support/nrsc5.py
```diff
@@ -754,9 +763,21 @@ def _callback_wrapper(self, c_evt):
                 c = comment_ptr.contents
                 comments.append(Comment(self._decode(c.lang), self._decode(c.short_content_desc), self._decode(c.full_text)))
                 comment_ptr = c.next
+            commercial = None
+            if id3.commercial.price:
```
argilo: The spec says that the price string may be empty, which would fail this check.
```suggestion
            if id3.commercial.price is not None:
```

include/nrsc5.h
```diff
@@ -453,6 +453,15 @@ struct nrsc5_event_t
...
             } xhdr;
             nrsc5_id3_comment_t *comments;
+            struct
+            {
+                char* price;            /** price of the product */
+                char* contact_url;      /** URL for contacting the seller */
+                char* seller;           /** Name of the seller */
+                char* description;      /** Short description of the product */
+                uint8_t received_as;    /** How audio is delivered when bought */
+                struct tm* valid_until; /** Date when the price expires */
```
```suggestion
                struct tm *valid_until; /** Date when the price expires. Note that only the `tm_year`, `tm_mon`, and `tm_mday` fields are set. */
```

include/nrsc5.h
```diff
@@ -453,6 +453,15 @@ struct nrsc5_event_t
                 int lot;
             } xhdr;
             nrsc5_id3_comment_t *comments;
+            struct
+            {
+                char* price;            /** price of the product */
+                char* contact_url;      /** URL for contacting the seller */
+                char* seller;           /** Name of the seller */
+                char* description;      /** Short description of the product */
```
```suggestion
                char *price;            /** price of the product */
                char *contact_url;      /** URL for contacting the seller */
                char *seller;           /** Name of the seller */
                char *description;      /** Short description of the product */
```

support/nrsc5.py
TheDaChicken: Wowz, that's an easy mistake to make. We may need to have better ways to check if fields exist for a specific type

PR #525, opened 2026-08-21
Reprocess SIG

### LINKED ISSUE #458: SIG table only processed once
libnrsc5 only processes the SIG table the first time it is seen:

https://github.com/theori-io/nrsc5/blob/20de3b2e9fdf2ca2a770bf7c38fb115ce4f7a998/src/output.c#L467-L471

This seems like a bad design, since a station could update its SIG table, or a different station's signal could arrive at the input. When changes to the SIG table occur and libnrsc5 ignores them, error messages like the following begin appearing:

```
port 0404 not defined in SIG table
port 0402 not defined in SIG table
sequence too large (-835829692)
sequence too large (136251704)
```

TheDaChicken:
Fixes #458 

Instead of only processing the SIG table once, this PR keeps processing the SIG table by finding the service number and updating its contents. The ordering is kept by swapping. All old services end up at the bottom

### DIFF
--- a/src/nrsc5.c
+++ b/src/nrsc5.c
@@ -2,6 +2,7 @@
 #include <string.h>
 
 #include "private.h"
+#include "unicode.h"
 
 pthread_mutex_t fftw_mutex = PTHREAD_MUTEX_INITIALIZER;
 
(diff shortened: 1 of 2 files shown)


argilo:
Closing, since this was replaced by #526.

PR #528, opened 2026-08-23
Fuzzy matching for PCI

### LINKED ISSUE #316: Tolerate bit errors in the PCI value
As noted in https://github.com/theori-io/nrsc5/issues/312#issuecomment-1624536654, the `has_fixed` function prevents data from being processed unless the PCI bits match one of the expected values exactly. When signal quality is poor, bits may be corrupted, disabling data decoding.

A couple possible solutions:

* Use fuzzy matching to determine the most likely PCI value.
* In the event of bit errors, assume that the PCI value is equal to the value seen in previous L2 PDUs.

TheDaChicken:
Fixes #316 
Reverts #315 

This PR adds fuzzy matching for PCI to correct for corrupted bits. This is helpful to ensure PDUs are decoded properly

### DIFF
--- a/src/frame.c
+++ b/src/frame.c
@@ -26,10 +26,11 @@
 #define PCI_AUDIO_FIXED 0xE3634C
 #define PCI_AUDIO_FIXED_OPP 0x8D8D33
 #define PCI_FIXED 0x3634CE
-#define PCI_RESERVED_1 0x8D338D
-#define PCI_RESERVED_2 0xD8D338
-#define PCI_RESERVED_3 0x634CE3
-#define PCI_COUNT 12
+#define PCI_RESERVED_CW5 0x8D338D
+#define PCI_RESERVED_CW6 0xD8D338
+#define PCI_RESERVED_CW7 0x634CE3
+#define PCI_MAX_ERRORS 4
+#define PCI_COUNT 8
 
 static const unsigned int PCI_POSSIBILITIES[PCI_COUNT] = {
     PCI_AUDIO,
@@ -37,9 +38,9 @@ static const unsigned int PCI_POSSIBILITIES[PCI_COUNT] = {
     PCI_AUDIO_FIXED,
     PCI_AUDIO_FIXED_OPP,
     PCI_FIXED,
-    PCI_RESERVED_1,
-    PCI_RESERVED_2,
-    PCI_RESERVED_3,
+    PCI_RESERVED_CW5,
+    PCI_RESERVED_CW6,
+    PCI_RESERVED_CW7,
 };
 
 #define MAX_AUDIO_PACKETS 64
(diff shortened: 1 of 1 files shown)


src/frame.c
```diff
@@ -26,6 +26,21 @@
 #define PCI_AUDIO_FIXED 0xE3634C
 #define PCI_AUDIO_FIXED_OPP 0x8D8D33
 #define PCI_FIXED 0x3634CE
+#define PCI_RESERVED_1 0x8D338D
+#define PCI_RESERVED_2 0xD8D338
+#define PCI_RESERVED_3 0x634CE3
+#define PCI_COUNT 12
```
argilo: This should be 8, shouldn't it?

TheDaChicken: Yep. You are correct. Silly me.

src/frame.c
```diff
@@ -26,6 +26,21 @@
...
+
+static const unsigned int PCI_POSSIBILITIES[PCI_COUNT] = {
+    PCI_AUDIO,
+    PCI_AUDIO_OPP,
+    PCI_AUDIO_FIXED,
+    PCI_AUDIO_FIXED_OPP,
+    PCI_FIXED,
+    PCI_RESERVED_1,
+    PCI_RESERVED_2,
+    PCI_RESERVED_3,
```
argilo: Maybe we could use the sequence names from table 5-6?
```suggestion
    PCI_RESERVED_CW5,
    PCI_RESERVED_CW6,
    PCI_RESERVED_CW7,
```

TheDaChicken: Not a bad idea actually. I agree.

src/frame.c
```diff
@@ -642,6 +660,19 @@ void frame_process(frame_t *st, size_t length, logical_channel_t lc)
 
 }
 
+static int fuzzy_pci(const unsigned int pci, const unsigned int pci_len, unsigned int* match)
+{
+    for (int i = 0; i < PCI_COUNT; i++)
+    {
+        const unsigned int score = __builtin_popcount((pci ^ PCI_POSSIBILITIES[i]) >> (24 - pci_len));
```
argilo: `__builtin_popcount` is a compiler-specific function. We should probably use a generic implementation for portability.

argilo: Performance is not a concern, since this is only called once per PDU.

src/frame.c
```diff
@@ -709,8 +749,18 @@ void frame_push(frame_t *st, uint8_t *bits, size_t length, logical_channel_t lc)
         }
     }
 
-    st->pci = header;
-    frame_process(st, ptr - st->buffer, lc);
+    const unsigned int len = ptr - st->buffer;
+    unsigned int pci = 0;
+    if (fuzzy_pci(header, pci_len, &pci) < 0)
+    {
+        if ((len == MAX_PDU_LEN || len == P1_PDU_LEN_AM))
```
argilo: I imagine this was copied over from `frame_process`. But I noticed it could be simplified:
```suggestion
        if (lc == P1_LOGICAL_CHANNEL)
```

argilo: Also, I wonder if we need to leave that desync code in `frame_process`. I'll see what happens if I pull it out.

argilo: I tried commenting out the desync in `frame_process` and got more audio packets decoded overall. Most low-SNR recordings had more audio packets decoded, but a few fared worse.

TheDaChicken: Are you saying I should remove the desync in frame_process?

argilo: I think so, but let me do a bit more testing to be sure.

TheDaChicken: Alright. I will wait on it to be sure.

argilo: I think it's the way to go. I also see more packet & stream data coming out when I remove the desync in `frame_process`.

src/frame.c
```diff
@@ -709,8 +744,18 @@ void frame_push(frame_t *st, uint8_t *bits, size_t length, logical_channel_t lc)
         }
     }
 
-    st->pci = header;
-    frame_process(st, ptr - st->buffer, lc);
+    const unsigned int len = ptr - st->buffer;
```
argilo: I suppose we don't need the `len` variable anymore.

src/frame.c
```diff
@@ -709,8 +744,18 @@ void frame_push(frame_t *st, uint8_t *bits, size_t length, logical_channel_t lc)
...
+    unsigned int pci = 0;
+    if (fuzzy_pci(header, pci_len, &pci) < 0)
```
argilo: Is there a reason not to assign the result directly to `st->pci`?

TheDaChicken: OH! TRUE! There's no reason not too. That's possible for me to do that.

PR #530, opened 2026-08-25
Add NAVTEQ traffic/alternate frequencies support

### LINKED ISSUE #418: Decode NavteqPacketData1 & NavteqAdmin
Stations with HERE TPEG & HERE Images also broadcast "NavteqPacketData1" and "NavteqAdmin" packet data. It appears these contain traffic data. It would be nice decode, or at least dump, this data.

billatq:
This change adds support for the NAVTEQ ALERT-C messages sent over HD Radio, and the alternate frequencies messages as well. There's also a script to convert a Garmin .TRF file into a .TSV so that it's straightforward to print out locations.

It's probably overkill to have so much ALERT-C parsing in there, but I have no idea if any of those messages are going to get used or not.

Resolves #418


billatq:
Updated the change to go through the normal logging functions versus outputting to stdout/stderr directly.

include/nrsc5.h
```diff
@@ -362,6 +364,48 @@ enum
...
+    uint16_t location;            /**< TMC location code. */
+    uint8_t direction;            /**< Raw direction bit. */
+    uint8_t extent;               /**< Number of adjacent locations affected. */
+    uint8_t bidirectional;        /**< Whether the event applies in both directions. */
+    uint8_t diversion;            /**< Whether drivers should follow a diversion. */
+    uint8_t duration_type;        /**< Encoded event duration type. */
+    uint8_t control_code;         /**< ALERT-C control code. */
+    uint16_t event;               /**< ALERT-C event code. */
+    uint8_t quantifier;           /**< Encoded ALERT-C quantifier; zero or 255 when absent. */
+} nrsc5_navteq_digital_traffic_entry_t;
```
TheDaChicken, line 383: `digital` seems redundant here. Other than that, I think the naming is fine. This also applies to `DIGITAL` in `NRSC5_EVENT_NAVTEQ_DIGITAL_TRAFFIC` and the collections for the Python API.

include/nrsc5.h
```diff
@@ -362,6 +364,48 @@ enum
...
+    uint8_t location_table_number;/**< TMC location table number. */
+    uint16_t location;            /**< TMC location code. */
+    uint8_t direction;            /**< Raw direction bit. */
+    uint8_t extent;               /**< Number of adjacent locations affected. */
+    uint8_t bidirectional;        /**< Whether the event applies in both directions. */
+    uint8_t diversion;            /**< Whether drivers should follow a diversion. */
+    uint8_t duration_type;        /**< Encoded event duration type. */
+    uint8_t control_code;         /**< ALERT-C control code. */
+    uint16_t event;               /**< ALERT-C event code. */
+    uint8_t quantifier;           /**< Encoded ALERT-C quantifier; zero or 255 when absent. */
```
TheDaChicken, line 382: For API clients, I think working with this would be easier if most of these fields weren't exposed. Instead a char pointer to the constructed message from the record type? 

In general though, @argilo Do you think that libnrsc5 should contain hard coded strings for each record type? Here's the included list: [alert_c_events.inc](https://github.com/theori-io/nrsc5/pull/530/changes#diff-6692614c04ad4a6ced395098ff3cf81f303967f58cd28b868c0a04785f24a2f7)

include/nrsc5.h
```diff
@@ -681,7 +745,31 @@ NRSC5_API void nrsc5_program_type_name(unsigned int type, const char **name);
...
+/** Opens an nrsc5 traffic location-table TSV file. Returns 0 on success or 1 on error. */
+NRSC5_API int nrsc5_location_table_open(nrsc5_location_table_t **result, const char *path);
+
+/** Closes a traffic location table. */
+NRSC5_API void nrsc5_location_table_close(nrsc5_location_table_t *table);
+
+/** Resolves a point. Returns 1 if found, 0 if absent, or -1 if malformed. */
+NRSC5_API int nrsc5_location_table_lookup(nrsc5_location_table_t *table,
+                                          uint8_t country, uint8_t ltn,
+                                          uint16_t location, nrsc5_location_t *result);
```
TheDaChicken, line 772: The location table lookup is also only used for nrsc5 clients. This seems like a lot to include? @argilo What do you think? 

While `garmin_trf_to_location_table.py` is super cool, couldn't anyone lookup locations in all sorts of ways ways from longitude and latitude?

src/main.c
```diff
@@ -1043,8 +1195,15 @@ int main(int argc, char *argv[])
 
     ao_initialize();
     init_audio_buffers(st);
-    if (parse_args(st, argc, argv) != 0)
-        return 0;
+    int parse_result = parse_args(st, argc, argv);
```
TheDaChicken, line 1198: While this is a problem with the C client, it might be good to include this in a different PR instead.

src/location_table.c
```diff
@@ -0,0 +1,308 @@
...
+static void destroy_numeric_locale(numeric_locale_t locale)
+{
+    freelocale(locale);
+}
+
+static double parse_locale_double(const char *text, char **end, numeric_locale_t locale)
+{
+    return strtod_l(text, end, locale);
+}
+#endif
```
TheDaChicken, line 78: From my understanding, North America uses only the regular Latin symbols for numbers. Since the Latin symbols are included in ASCII, the simple `atoi` function should be fine

README.md
```diff
@@ -162,3 +164,43 @@ If you get errors trying to access your RTL-SDR device, then you may need to use
...
+
+Then enable debug logging and load the table to print location names,
+coordinates, extent endpoints, and descriptions for Garmin Digital Traffic
+events:
+
+```sh
+nrsc5 -l 1 --location-table locations.tsv 97.1 0
+```
+
+The generated table contains one point per line with country, location-table
```
TheDaChicken, line 204: I don't think this description is needed here. This could be in your PR description.

src/CMakeLists.txt
```diff
@@ -16,13 +16,13 @@ if (USE_STATIC)
...
 
 configure_file (config.h.in config.h)
+include_directories (BEFORE "${CMAKE_SOURCE_DIR}/include")
 include_directories (
     ${FAAD2_INCLUDE_DIRS}
     ${AO_INCLUDE_DIRS}
     ${RTL_SDR_INCLUDE_DIRS}
     ${FFTW3F_INCLUDE_DIRS}
     "${CMAKE_CURRENT_BINARY_DIR}"
-    "${CMAKE_SOURCE_DIR}/include"
```
TheDaChicken, line 25: I’m not sure this change is necessary.

include/nrsc5.h
TheDaChicken: `format_ordinal` and `format_quantifier` seem to also be similar. A specific ID to provide custom formatting. This could be up to the API user on how they might describe the time etc.

include/nrsc5.h
billatq: Unfortunately you don't get lon/lat on the wire, you get an opaque location reference, say something like 12345. You need to do this lookup to get the coordinates. Depending on how many locations you get, it may be possible you work it out, but my local station emits around 13k of these at a time.

src/main.c
billatq: Will move it out of this PR.

include/nrsc5.h
TheDaChicken: > Unfortunately you don't get lon/lat on the wire

Over the air you mean?

Good to know. I understand.

README.md
TheDaChicken: Some of it may be required depending on what happens.

include/nrsc5.h
billatq: In the frame, but "You see, wire telegraph is a kind of a very, very long cat. You pull his tail in New York and his head is meowing in Los Angeles. Do you understand this? And radio operates exactly the same way: you send signals here, they receive them there. The only difference is that there is no cat." (or wire) :)

TheDaChicken: Yes, "in the frame" makes sense. I dont understand the rest of your sentence.

Issue #471, opened 2025-12-23
Replace libao with miniaudio


pclov3r:
https://github.com/theori-io/nrsc5/issues/463 and https://github.com/theori-io/nrsc5/issues/334 show that perhaps libao has some issues. 

Given it seems to be almost abandoned now perhaps it might be good to replace it with miniaudio. This actually drops a build dep requirement as miniuadio is a single header library and has wav encoding built in.

pclov3r:
miniaudio is a single header library you don't need to install anything or build a separate library. 

https://github.com/mackron/miniaudio

> Just compile miniaudio.c like any other source file and include miniaudio.h like a normal header. There's no need to install any dependencies. On Windows and macOS there's no need to link to anything. On Linux and BSD just link to -lpthread and -lm. On iOS you need to compile as Objective-C. Link to -ldl if you get errors about dlopen(), etc.

Issue #334, opened 2023-11-20
Sporadic audio output issues ...


gstalnaker:
Linux HOLLIN 6.1.0-10-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.38-2 (2023-07-27) x86_64 GNU/Linux

MX Linux MX-23.1, KDE Plasma 5.27.5, Debian 12.2

nrsc5 revision 80ab64e, RTL-SDR v3 receiver, using the dipole attenna.

### Short Story

Sometimes nrsc5 runs, appears to connect to the RTL-SDR dongle, and start playing the streams, only to have the audio stop playing.

### Long Story

I Run nrsc5 at a terminal prompt:

```
*[master][~/bin/nrsc5-dui]$ /usr/local/bin/nrsc5 -q 104.1 1
Found Rafael Micro R820T tuner
Exact sample rate is: 1488375.071248 Hz
[R82XX] PLL not locked!
Allocating 8 zero-copy buffers
```

What I hear is a small blip, chirp, 5-10 seconds in, and then nothing. This happens both running nrsc5 from a terminal prompt and from within nrsc5-dui.

I run with logging enabled:

```
*[master][~/bin/nrsc5-dui]$ /usr/local/bin/nrsc5 -l 1 88.7 0
Found Rafael Micro R820T tuner
Exact sample rate is: 1488375.071248 Hz
[R82XX] PLL not locked!
Allocating 8 zero-copy buffers
21:22:47 Synchronized
21:22:47 Country: US, FCC facility ID: 63030
21:22:47 Station name: WERN-FM
21:22:47 Country: US, FCC facility ID: 63030
21:22:47 Station name: WERN-FM
21:22:47 Audio program 1: public, type: Public, sound experience 0
21:22:47 Audio program 2: public, type: Talk, sound experience 0
21:22:47 Country: US, FCC facility ID: 63030
21:22:47 Station name: WERN-FM
21:22:47 Slogan: MPS
21:22:47 Audio program 1: public, type: Public, sound experience 0
21:22:47 Audio program 2: public, type: Talk, sound experience 0
21:22:48 Country: US, FCC facility ID: 63030
21:22:48 Station name: WERN-FM
21:22:48 Slogan: MPS
21:22:48 Audio program 0: public, type: Public, sound experience 0
21:22:48 Audio program 1: public, type: Public, sound experience 0
21:22:48 Audio program 2: public, type: Talk, sound experience 0
[...]

PR #532, opened 2026-09-01
Replace libao with miniaudio.

TheDaChicken:
#471 

Libao has some issues which affects using the C client like #334 . To hopefully fix these issues, I replaced libao with miniaudio. In my testing on my Windows machine, I noticed audio catches up a lot better when my computer temporarily freezes.

Miniaudio is a simple header-only library. I included the header directly in this PR (this explains the huge size of the PR). If preferred, I can code up CMake to automatically download this header. I would prefer to avoid modifying CMake (I don't like CMake).

Miniaudio outputs wav files with RF64 headers. This increases the size limit for wav files from 4GBs to 16 exabytes. This is helpful for those using wav output to a audio player or such that is running 24/7. 

For this PR, I used miniaudio's thread-safe lock free buffer. Similar to PortAudio, it's recommended to use a lock-free audio buffer. Audio callbacks are usually ran on the OS thread. Locking on the OS thread isn't good practice. This means that I had to change up how IQ input works since `push_audio_buffer` doesn't block anymore. I merged the IQ input thread with the input thread since both are now non-blocking.


src/main.c
```diff
@@ -36,29 +35,32 @@
...
     IQ_FORMAT_CU8,
     IQ_FORMAT_CS16,
     IQ_FORMAT_CF32,
 };
 
+enum output_format
+{
+    WAV,
+    RAW,
+    DEVICE,
```
argilo: C enums are not namespaced, so it would probably be worth prefixing the names like so:
```suggestion
    OUTPUT_FORMAT_WAV,
    OUTPUT_FORMAT_RAW,
    OUTPUT_FORMAT_DEVICE,
```

TheDaChicken: Ah! Got it.

argilo:
Libao is still referenced in several places:

* CMakeLists.txt: `${AO_BUILTIN}`
* src/CMakeLists.txt: `${AO_INCLUDE_DIRS}`, `${AO_LIBRARY_DIRS}`
* .github/workflows/docs.yml: `libao-dev`
* support/msys2-build: `-D USE_SYSTEM_LIBAO=OFF`
* support/win-cross-compile: `-D USE_SYSTEM_LIBAO=OFF`

TheDaChicken:
Oops, all references should be completely removed now!

argilo:
> Miniaudio outputs wav files with RF64 headers.

Does it? I don't see it mentioned in the miniaudio documentation, and didn't get RF64 output when I fed in a large IQ recording.

TheDaChicken:
> didn't get RF64 output when I fed in a large

Oh, I thought rf64 was the default in the library. On line 87883, it's possible to change `ma_dr_wav_container_riff` to `ma_dr_wav_container_rf64`. I could switch out & directly use the wav encoder that's integrated in the library so it's possible to customize this option

argilo:
As far as I could tell, the RF64 support is only for reading, not writing.

TheDaChicken:
Right here it describes support rf64:
https://github.com/mackron/dr_libs/blob/master/dr_wav.h#L65

I tried changing that line I suggested just now and saw "RF64" at the beginning of the outputted file.

TheDaChicken:
Well, no it doesn't directly say the library supports r64 in the link i gave. The comments describes the library doesn't support "AIFF or RIFX". I assume rf64 is supported & works. 

EDIT: Found a better link describing what's supported: https://github.com/mackron/dr_libs/blob/dfe8377631000664666519fdb83da193fd8037f4/dr_wav.h#L101

TheDaChicken:
Sadly, I couldn't use the internal dr wav functions. Drwav's function declarations were included in the miniaudio implementation. Since including method definitions would slow down the compilation of main.c each time, I added drwav.h and drwav.c. 

This PR now does, in fact, output WAVs at rf64! :)

Issue #489, opened 2026-04-12
No API indication of silence inserted audio


TheDaChicken:
Currently, there is no way for API clients to know when silence was inserted other than crc events for hdc packets.

There are currently two other cases that can cause silence to be inserted:

- When the audio decoder is still starting up, silence must be inserted 
- When the audio decoder fails to decode.

While the last one is rare, it would be nice to tell the user in a rare event that errors happen.

I propose maybe a flag for the audio event that can provide this information to the api client.

argilo:
This feature was implemented in #533.

PR #533, opened 2026-09-07
Add Audio Flags for decoding failed & audio silence

TheDaChicken:
Adds #489 / #372

Since libnrsc5 defaults to no logging, there's no indication for audio errors if there are. This adds them to the API to at least give give an indication of an error to the C client user. 

It may be helpful for the API user to also display on their UI if silence was inserted. This PR also adds that to the API. This is defined as `NRSC5_AUDIO_FLAGS_UNAVAILABLE` (defined as when digital audio is unavailable) to hopefully make this more future-proof if any analog blending is added.

### DIFF
--- a/include/nrsc5.h
+++ b/include/nrsc5.h
@@ -362,6 +362,13 @@ enum
     NRSC5_PKT_FLAGS_CRC_ERROR = 1 << 0, /** Failed the CRC check. Could be corrupted packet. */
 };
 
+enum
+{
+    NRSC5_AUDIO_FLAGS_NONE = 0,
+    NRSC5_AUDIO_FLAGS_UNAVAILABLE = 1 << 0 /** Digital audio was currently unavailable. */,
+    NRSC5_AUDIO_FLAGS_DECODING_ERROR = 2 << 0, /** Failed audio decoding. */
+};
+
 /**  Incoming event from receiver.
  *
  * This event structure is passed to your application supplied
@@ -436,6 +443,7 @@ struct nrsc5_event_t
             unsigned int program;
             const int16_t *data;
             size_t count;
+            unsigned int flags; /** The specific status of the audio sample. Example `NRSC5_AUDIO_FLAGS_DECODING_ERROR` **/
         } audio;
         struct {
             unsigned int program;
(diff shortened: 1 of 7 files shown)


src/main.c
```diff
@@ -367,6 +367,8 @@ static void callback(const nrsc5_event_t *evt, void *opaque)
         }
         break;
     case NRSC5_EVENT_AUDIO:
+        if (evt->audio.flags & NRSC5_AUDIO_FLAGS_DECODING_ERROR)
+            log_error("Audio decoding error");
```
argilo: I think warn makes sense here, since decoding errors are expected to occur occasionally when SNR is low.
```suggestion
            log_warn("Audio decoding error");
```

argilo:
The Python API and client should be updated with the same changes that have been made in the C versions.

TheDaChicken:
Oh. Also do you think I should always print out when there's a decoding error or should I print it out only for the current program?

argilo:
I think it would be best to display it for the current program only, like we do for CRC errors.

include/nrsc5.h
```diff
@@ -362,6 +362,13 @@ enum
     NRSC5_PKT_FLAGS_CRC_ERROR = 1 << 0, /** Failed the CRC check. Could be corrupted packet. */
 };
 
+enum
+{
+    NRSC5_AUDIO_FLAGS_NONE = 0,
+    NRSC5_AUDIO_FLAGS_UNAVAILABLE = 1 << 0 /** Digital audio was currently unavailable. */,
+    NRSC5_AUDIO_FLAGS_DECODING_ERROR = 2 << 0, /** Failed audio decoding. */
```
```suggestion
    NRSC5_AUDIO_FLAGS_DECODING_ERROR = 1 << 1, /** Failed audio decoding. */
```

support/nrsc5.py
```diff
@@ -194,13 +194,18 @@ class PacketFlags(enum.IntFlag):
     NONE = 0
     CRC_ERROR = 1 << 0
 
+class AudioFlags(enum.IntFlag):
+    NONE = 0
+    UNAVAILABLE = 1 << 0
+    DECODING_ERROR = 2 << 0
```
argilo:
```suggestion
    DECODING_ERROR = 1 << 1
```

